PR #5, opened 2020-03-05
images for rpm queries and caching

sosiouxme:


artbotlib/brew_list.py
```diff
@@ -0,0 +1,173 @@
+import json
+import koji
+
+from . import util
+
+# TODO: use a real cache with expiry and stuff
```
jupierce: https://pypi.org/project/cachetools/ ?

PR #7, opened 2020-03-22
config file and containerization

sosiouxme:


art-bot.py
```diff
@@ -251,57 +235,58 @@ def consumer_thread(client_id, topic, callback_handler, consumer, durable, user_
...
+    except Exception as exc:
+        print(f"Error loading art-bot config file {config_file}: {exc}")
+        exit(1)
+
+    def abs_path_home(filename):
+        # if not absolute, relative to home dir
+        return filename if filename.startswith("/") else f"{os.environ['HOME']}/{filename}"
+
+    try:
+        with open(abs_path_home(bot_config["slack_api_token_file"]), "r") as stream:
```
jupierce: Unless you firmly want this, I would vote to keep this an environment variable only. If it is in a file, it will eventually be added to a commit my someone. A single push and we will be scrambling.

sosiouxme: @jupierce does it help that it's typically going to be a file in `~/.config/art-bot/` i.e. away from the repo?

I get the concern, but I imagine someone's as likely to put the env var definition in a file and commit that somewhere.

jupierce: Yes. I missed the ~/.config/art-bot. That does help.

PR #11, opened 2021-03-12
Fix chunk response

vfreex:


sosiouxme (approved):
i had to test out some things, sorry for the delay... looks good, thanks!

PR #15, opened 2022-05-31
Minor updates

locriandev:
- add Pycharm files to ignored patterns
- use :base tag when building base image, as it's used as such [here](https://github.com/openshift/art-bot/blob/master/container/Dockerfile.latest#L1)
- update krb5-redhat.conf (the old one was no longer working)

### DIFF
--- a/container/README.md
+++ b/container/README.md
@@ -3,7 +3,7 @@
 ## Production
 
 Normally art-bot runs as an OpenShift workload in our ocp4 space (see
-https://github.com/openshift-eng/art-docs/blob/master/infra/art-bot.md). The
+https://github.com/openshift-eng/art-docs/blob/master/content/infra/art-bot.md). The
 container is built initially from `Dockerfile` and then code-only updates can
 be added with `Dockerfile.latest` (skipping all the RPM installs etc).
 Credentials and config are mounted in as secrets and configmaps.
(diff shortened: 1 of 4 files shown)


container/README.md
```diff
@@ -56,7 +56,7 @@ To simplify these steps, here is a script that takes a parameter to build these:
 base_cmd="podman build --build-arg USERNAME=lmeyer --build-arg USER_UID=3668"
 
 if [[ "$1" == base ]]; then
-    $base_cmd -f container/Dockerfile -t art-bot:latest .
+    $base_cmd -f container/Dockerfile -t art-bot:base .
```
sosiouxme, line 59: don't think i want this change; the idea is that you run this once (to get dependencies from scratch), and then for code updates you can just add a layer to update the code. so both are going to `:latest`
```suggestion
    $base_cmd -f container/Dockerfile -t art-bot:latest .
```

container/README.md
```diff
@@ -3,7 +3,7 @@
 ## Production
 
 Normally art-bot runs as an OpenShift workload in our ocp4 space (see
-https://github.com/openshift-eng/art-docs/blob/master/infra/art-bot.md). The
+https://github.com/openshift-eng/art-docs/blob/master/content/infra/art-bot.md). The
```
sosiouxme, line 6: thanks for pointing out it's moved. but how about just linking to the doc site :) 
```suggestion
https://art-docs.engineering.redhat.com/infra/art-bot/). The
```

PR #16, opened 2022-06-09
Distgit to Delivery Repo mappings

ashwindasr:
Map Distgit repo name to Delivery repo name starting from the former.


artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,97 @@
...
+            print("ERROR: Authentication failure. Kerberos ticket")
+
+        return response.json()['data']['attributes']['external_name']
+    except Exception as e:
+        print(f"Error: {e}")
+        return "Couldn't find delivery repo name."
+
+
+def check_distgit_availability(distgit_repo_name):
+    response = requests.get(f"https://pkgs.devel.redhat.com/cgit/containers/{distgit_repo_name}")
```
sosiouxme, line 64: if you just need the status code then a `.head` request is quicker/kinder (doesn't have to generate the page)

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    List the Brew package name, CDN repo name and CDN repo details by getting the distgit name as input.
+
+    :so: SlackOuput object for reporting results.
+    :distgit_repo_name: Name of the distgit repo we get as input
+    :version: OS version
+    """
+
+    if not version:
+        version = "4.11"  # Default version set to 4.11, if unspecified
```
sosiouxme, line 79: should be 4.10 for now since we've shipped that and CDN repos are configured

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,97 @@
...
+
+    payload = ""
+    if check_distgit_availability(distgit_repo_name): # Check if the given distgit repo actually exists
+        payload += f"Distgit Repo: <https://pkgs.devel.redhat.com/cgit/containers/{distgit_repo_name}|*{distgit_repo_name}*>\n"
+
+        brew_package_name = distgit_to_brew(distgit_repo_name, version)
+        cdn_repo_name = brew_to_cdn(brew_package_name, "8Base-RHOSE-4.10")  # Default variant set to 8Base-RHOSE-4.10"
```
sosiouxme, line 86:
```suggestion
        cdn_repo_name = brew_to_cdn(brew_package_name, f"8Base-RHOSE-{version}")
```

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,97 @@
...
+        # Sending the kerberos ticket along with the request
+        response = requests.get(
+            f"https://errata.devel.redhat.com/api/v1/cdn_repo_package_tags?page[number]=1&filter[package_name]={brew_name}&filter[variant_name]={variant_name}",
+            auth=kerberos_auth)
+
+        if response.status_code == 401:
+            print("ERROR: Authentication failure. Enter kerberos password for ticket")
+
+        json_file = response.json()
+        return json_file['data'][0]['relationships']['cdn_repo']['name']
```
sosiouxme, line 39: i suspect there can be more than one entry in `data` but i don't have a ready example to see... note to self to try

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,97 @@
...
+    except Exception as e:
+        print(f"Error: {e}")
+        return "Couldn't find CDN repo name."
```
sosiouxme, line 42: again would prefer an exception here, with enough information for the caller to be able to tell the user "errata-tool says there's no CDN repo for that in version {version}" or "something unexpected went wrong querying for CDN repo" and write something to `art-bot-monitoring`

and not having a CDN repo is fine, we just can't get past that to find comet repo so we have to tell the user :shrug:

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,97 @@
...
+        if response.status_code == 401:
+            print("ERROR: Authentication failure. Kerberos ticket")
+
+        return response.json()['data']['attributes']['external_name']
+    except Exception as e:
+        print(f"Error: {e}")
+        return "Couldn't find delivery repo name."
+
+
+def check_distgit_availability(distgit_repo_name):
```
sosiouxme, line 63: since you're using it like a boolean i'd suggest
```suggestion
def distgit_is_available(distgit_repo_name):
```
or maybe
```suggestion
def distgit_exists(distgit_repo_name):
```

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,97 @@
...
+
+        brew_package_name = distgit_to_brew(distgit_repo_name, version)
+        cdn_repo_name = brew_to_cdn(brew_package_name, "8Base-RHOSE-4.10")  # Default variant set to 8Base-RHOSE-4.10"
+        cdn_repo_id = cdn_to_comet(cdn_repo_name)
+
+        payload += f"Brew package: *{brew_package_name}*\n"
+        payload += f"CDN repo: *{cdn_repo_name}*\n"
+        payload += f"Delivery (Comet) repo: *{cdn_repo_id}*\n"
+
+    else:
```
sosiouxme, line 93: i did not seem to reach this branch of logic when I requested a bogus repo. i don't see why not though.

tests/test_pipeline_image_names.py
```diff
@@ -0,0 +1,57 @@
+from artbotlib.pipeline_image_names import distgit_to_brew, brew_to_cdn, cdn_to_comet, check_distgit_availability
+
+
+def test_distgit_to_brew_1():
```
sosiouxme, line 4: nice to see tests.
technically these are functional/integration tests, since they'll hit the live system for info. so they should go in `functional`. you could also mock out the system calls in order to make them actual unit tests...

sosiouxme (changes requested):
a promising start; some things to try/discuss

artbotlib/pipeline_image_names.py
ashwindasr: Done!

artbotlib/pipeline_image_names.py
ashwindasr: Done!

artbotlib/pipeline_image_names.py
ashwindasr: Done!

artbotlib/pipeline_image_names.py
ashwindasr: I added that change with 49d94af17a73eb62be4c849d0df5e9d9c0270fba. Can you check to see if its working again?

image

tests/test_pipeline_image_names.py
ashwindasr: Done!

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,97 @@
...
+        component_name = yml_file['distgit']['component']
+    except KeyError:
+        return brew_name
+
+    return component_name
+
+
+def brew_to_cdn(brew_name, variant_name):
+    # Kerberos authentication
+    kerberos_auth = HTTPKerberosAuth(mutual_authentication=REQUIRED, sanitize_mutual_error_response=False)
```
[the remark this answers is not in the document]
ashwindasr: Okay! So no need to make any changes right?

artbotlib/pipeline_image_names.py
ashwindasr: You're right. So the API checks only the first page. `openshift-enterprise-console-container` has only one instance in all the pages though. But like you said there could be others in other pages right.  Should I go through all the data and check if there's any other one that has `package_name` as the brew package name and display multiple outputs or ?

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,97 @@
...
+
+def distgit_to_brew(distgit_name, version):
+    brew_name = f"{distgit_name}-container"
+
+    response = requests.get(
+        f"https://raw.githubusercontent.com/openshift/ocp-build-data/openshift-{version}/images/{distgit_name}.yml")
+
+    if response.status_code != 200:
+        print(f"Error: The specified yml file doesn't exist")
+        return "Couldn't find Brew Package name"
```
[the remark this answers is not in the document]
ashwindasr: Done!

artbotlib/pipeline_image_names.py
ashwindasr: Okay! I added the custom exception, please check to see if that's okay. Also just to confirm, the function to display to art-bot-monitoring is `so.monitoring_say()` ?

ashwindasr:
Also need to add links for Brew, CDN and Delivery repos.

artbotlib/pipeline_image_names.py
sosiouxme: looks good now, must not have had this update :+1:

ashwindasr:
Finished adding links to all parts of the pipeline.

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,422 @@
...
+        if array != ['']:
+            dict_data[array[1]] = array[0]
+    return dict_data
+
+
+def distgit_github_mappings(version):
+    output = util.cmd_gather("doozer -g openshift-" + version + " images:print --short '{name}: {upstream_public}'")
+    data = output[1].split("\n")
+    dict_data = {}
+    for line in data:
```
sosiouxme, line 335: there's a pattern for this...
```suggestion
    dict_data = {}
    for line in output[1].splitlines():
```

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,422 @@
...
+        array = line.split(": ")
+        if array != ['']:
```
sosiouxme, line 337: seems overly specific. this shouldn't happen with this output and `splitlines()`, but if it did, i'd probably want to ignore more:
```suggestion
        if len(array) == 2:
```

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,422 @@
...
+    return dict_data
+
+
+@util.cached
+def get_github_from_distgit(distgit_name, version):
+    data = distgit_github_mappings(version)
+    try:
+        return data[distgit_name].split('/')[-1]
+    except Exception:
+        raise GithubFromDistgitNotFound(distgit_name, version)
```
sosiouxme, line 348: this doesn't seem to be getting caught and notified back to user. trying
`What is the image pipeline for distgit aos3-installation with version 3.11`
... just crashes. (Also it should be found... not sure why it's not)

sosiouxme (changes requested):
It would be surprising if I didn't find anything to tweak :)

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,422 @@
...
+            so.say(e.message)
+            so.monitoring_say(f"ERROR: {e.message}")
+        except Exception as e:
+            so.say("Unknown error. Contact the ART team.")
+            so.monitoring_say(f"ERROR: Unclassified: {e}")
+
+    else:
+        # If incorrect distgit name provided, no need to proceed.
+        payload += f"No distgit repo with name *{distgit_repo_name}* exists."
+        so.monitoring_say(f"No distgit repo with name *{distgit_repo_name}* exists.")
```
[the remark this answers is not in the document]
ashwindasr: Inverted logic and removed error logs to monitoring channel

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,422 @@
...
+
+    return component_name
+
+
+def get_image_stream_tag(distgit_name, version):
+    url = f"https://raw.githubusercontent.com/openshift/ocp-build-data/openshift-{version}/images/{distgit_name}.yml"
+    response = requests.get(url)
+
+    yml_file = yaml.safe_load(response.content)
+    if yml_file['for_payload']:
```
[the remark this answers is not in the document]
ashwindasr: Okay!

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,422 @@
...
+        tag = yml_file['name'].split("/")[1]
+        if "ose" in tag:
+            return tag[4:]
+        else:
+            return tag
```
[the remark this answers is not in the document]
ashwindasr: Understood!

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,422 @@
...
+    dict_data = {}
+    for line in data:
+        array = line.split(": ")
+        if array != ['']:
+            dict_data[array[0]] = array[1]
+    return dict_data
+
+
+@util.cached
+def get_github_from_distgit(distgit_name, version):
```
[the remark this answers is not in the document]
ashwindasr: Makes sense!

artbotlib/pipeline_image_names.py
ashwindasr: Ohh I see! Thank you!

artbotlib/pipeline_image_names.py
ashwindasr: Yeah I found the issue as well. It was because it was outside the `try` block. Fixed!

ashwindasr: And regarding why it's not working, it seems to be a doozer issue:
```bash
    raise ValueError(f'{self.distgit_key} claims to be for_release and base_only')
ValueError: openshift-enterprise-base claims to be for_release and base_only
```
for the command `doozer -g openshift-3.11 images:print --short '{name}: {upstream_public}'`

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,422 @@
...
+        return data[distgit_name].split('/')[-1]
+    except Exception:
+        raise GithubFromDistgitNotFound(distgit_name, version)
+
+
+def get_distgit_from_github(git_repo, version):
+    data = github_distgit_mappings(version)
+    return data[git_repo]
+    # TODO: For next iteration
+
```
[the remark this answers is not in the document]
ashwindasr: Understood. I forgot to clean it before I committed, my bad

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,422 @@
...
+    Attributes:
+        distgit_name -- input dist-git name which caused the error
+        message -- explanation of the error
+    """
+
+    def __init__(self, distgit_name, url):
+        self.distgit_name = distgit_name
+        self.message = f"image dist-git {distgit_name} definition was not found at {url}"
+        super().__init__(self.message)
```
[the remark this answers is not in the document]
ashwindasr: Done!

ashwindasr:
Rebuilt the art-bot container, but `click` version `8.1.3` gets installed instead of `8.1.2`. Found that the version is being pulled from the `requirements.txt` of `art-bot`. Adding a commit to lock it at `8.1.2`

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,265 @@
...
+
+# Driver functions
+def pipeline_from_github(so, github_repo, version):
+    """
+    Function to list the GitHub repo, Brew package name, CDN repo name and delivery repo by getting the GitHub repo name as input.
+
+    GitHub -> Distgit -> Brew -> CDN -> Delivery
+
+    :so: SlackOutput object for reporting results.
+    :github_repo: Name of the GitHub repo we get as input
```
sosiouxme, line 13: This is optional, but in general if there's more than one obvious way to refer to something, I like to accept any reasonable way.

For this case, I'd consider any of the following equivalent and wasn't sure which to use:
* `ironic-image`
* `openshift/ironic-image`
* `github.com/openshift/ironic-image`
* `https://github.com/openshift/ironic-image.git`
* `https://github.com/openshift/ironic-image/`
* `https://github.com/openshift/ironic-image`

I'd like to see whatever comes in get normalized to the last form and then proceed from that.

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,265 @@
...
+    """
+    if not version:
+        version = "4.10"  # Default version set to 4.10, if unspecified
+    variant = f"8Base-RHOSE-{version}"
+
+    payload = ""
+
+    if not pipeline_image_util.github_repo_is_available(github_repo):  # Check if the given GitHub repo actually exists
+        # If incorrect GitHub name provided, no need to proceed.
+        payload += f"No GitHub repo with name *{github_repo}* exists."
```
sosiouxme, line 24: It is helpful here to give an example of what it should look like, in case they just didn't know what format to use.

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def pipeline_from_distgit(so, distgit_repo_name, version):
+    """
+    Function to list the GitHub repo, Brew package name, CDN repo name and delivery repo by getting the distgit name as input.
+
+    GitHub <- Distgit -> Brew -> CDN -> Delivery
+
+    :so: SlackOutput object for reporting results.
+    :distgit_repo_name: Name of the distgit repo we get as input
```
sosiouxme, line 69: Likewise here it would be nice to accept `containers/ironic` which is technically the repo, or `ironic`

(and/or give an example when they get it wrong)

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,265 @@
...
+    if not pipeline_image_util.brew_is_available(brew_name):  # Check if the given brew repo actually exists
+        # If incorrect brew name provided, no need to proceed.
+        payload += f"No brew repo with name *{brew_name}* exists."
+        so.say(payload)
+        return
+    else:
+        so.say("Fetching data. Please wait...")
+        try:
+            # Brew -> GitHub
+            payload += pipeline_image_util.brew_to_github(brew_name, version)
```
sosiouxme, line 138: There is a bug here somewhere. This got stuck on the way to github:
```
sosiouxme: @art-bot-luke pipeline for package ironic-container in 4.11
art-bot-luke: Fetching data. Please wait...
art-bot-luke: Couldn't find GitHub repo from distgit ironic and version 4.11
```
`ironic` is the correct distgit but it did not map back correctly to GitHub.

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,265 @@
...
+    """
+    if not version:
+        version = "4.10"  # Default version set to 4.10, if unspecified
+    variant = f"8Base-RHOSE-{version}"
+
+    payload = ""
+
+    if not pipeline_image_util.cdn_is_available(cdn_repo_name):  # Check if the given brew repo actually exists
+        # If incorrect brew name provided, no need to proceed.
+        payload += f"No CDN repo with name *{cdn_repo_name}* exists."
```
sosiouxme, line 181: no one is really likely to search by this, but if they do... best to give them an example here of what a CDN repo even looks like (they likely don't know)

artbotlib/pipeline_image_names.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def pipeline_from_delivery(so, delivery_repo_name, version):
+    """
+    Function to list the GitHub repo, Brew package name, CDN repo name and delivery repo by getting the delivery repo name as input.
+
+    GitHub <- Distgit <- Brew <- CDN <- Delivery
+
+    :so: SlackOutput object for reporting results.
+    :delivery_repo_name: Name of the delivery repo we get as input
```
sosiouxme, line 218: again optional, but I'd expect to accept any of:
* `registry.redhat.io/openshift4/ose-ironic-rhel8`
* `openshift4/ose-ironic-rhel8`
* `ose-ironic-rhel8`

... and/or an example of the expected format

sosiouxme (changes requested):
other than one bug found, looking pretty good. going to push some cosmetic changes as well as a fix for a totally unrelated bug i found

artbotlib/pipeline_image_names.py
ashwindasr: Found that the issue was with `cache`.

artbotlib/pipeline_image_names.py
ashwindasr: I wrote a regex to match all the above cases, but there seems to be an issue with getting an URL input for the last 3 cases. Slack is sending the data like this:
```
----------------- DATA -----------------

{'blocks': [{'block_id': '8eFd',
             'elements': [{'elements': [{'text': '@art-bot-dev what is the '
                                                 'image pipeline for github ',
                                         'type': 'text'},
                                        {'style': {'unlink': True},
                                         'text': 'https://github.com/openshift/ironic-image.git',
                                         'type': 'text'}],
                           'type': 'rich_text_section'}],
             'type': 'rich_text'}],
```
So its not getting matched with the regex as URL is being sent differently than the text. It works if I manually remove the hyperlink and send the request. Normally the app reads it like this
```
----------------- DATA -----------------

{'channel': 'D03F6UXS5N3',
 'event_ts': '1657135302.007400',
 'hidden': True,
 'message': {'blocks': [{'block_id': 'hjD',
                         'elements': [{'elements': [{'text': '@art-bot-dev '
                                                             'what is the '
                                                             'image pipeline '
                                                             'for distgit '
                                                             'openshift-enterprise-cli',
                                                     'type': 'text'}],
                                       'type': 'rich_text_section'}],
                         'type': 'rich_text'}],
```

ashwindasr: I wrote a regex to match all the above cases: ```(https:\/\/)*(github.com\/)*(openshift\/)*([a-zA-Z0-9-]+)(.*)```
But when talk to artbot like ```@art-bot-dev what is the image pipeline for github https://github.com/openshift/ironic-image.git```, Slack it converting the URL to hyperlink and I'm getting ```None``` in the backend. It works when I change the URL to plain text in slack though, but I'm not sure if people will actively do that

artbotlib/pipeline_image_names.py
ashwindasr: Same hyperlink issue in the first example.

artbotlib/pipeline_image_names.py
ashwindasr: image

artbotlib/pipeline_image_names.py
sosiouxme: https://github.com/openshift/ocp-build-data/commit/15b40758b5f95627c40a85134453ab13bb451b07 should fix that.

sosiouxme: technically doozer shouldn't have two attrs that are basically for the same thing but... history :(

artbotlib/pipeline_image_names.py
sosiouxme: add in https://github.com/openshift/art-bot/pull/20 which should cause the url to show up in the query

PR #17, opened 2022-06-13
Fix: handle arm64 buildinfo

locriandev:


sosiouxme (approved):
/lgtm

i mean a unit test would be nice but...

PR #18, opened 2022-07-08
Add kernel info capability

locriandev:


ashwindasr (commented):
Tested it against RCs and its working :heavy_check_mark: 
When I tried with a release ```@art-bot-dev what kernel is used in 44.10.26``` I got this error though:
```
error: image "quay.io/openshift-release-dev/ocp-release:44.10.26-x86_64" not found: manifest unknown: manifest unknown
```
I'm not sure if its because I don't have enough authorization though.

artbotlib/buildinfo.py
```diff
@@ -122,3 +164,65 @@ def brew_build_url(nvr):
...
+    async def gather_kernel_info():
+        tasks = [
+            asyncio.ensure_future(non_rhcos_kernel_info('driver-toolkit')),
+            asyncio.ensure_future(non_rhcos_kernel_info('ironic-machine-os-downloader')),
+            asyncio.ensure_future(rhcos_kernel_info())
+        ]
+        return await asyncio.gather(*tasks)
+
+    # Format output and send to Slack
+    res = loop.run_until_complete(gather_kernel_info())
```
ashwindasr, line 221: Also it might be a good idea to catch an error if something goes wrong in one of the `oc` commands and end the `asyncio event loop`. Otherwise will the loop will wait around infinitely right (or until it times out). When I got the auth error, the tasks were still running until I closed it.

ashwindasr, line 221: Also does retrieving the data here take a lot amount of time? When I ran the job it finished quickly. If it something that doesn't take much time, we can avoid `async` and just call them normally no?

locriandev:
> @art-bot-dev what kernel is used in 44.10.26

I think you have mispelled the release nvr, trying with `@art-bot-dev what kernel is used  in 4.10.26` work fine. I  should certainly handle this error though, since it seems like art-bot does not reply anything even if the error is shown in the logs

artbotlib/buildinfo.py
locriandev: `asyncio` here is very important to run brew queries and `oc` commands in parallel. Each single task can take quite a few seconds, running them sequentially would take an annoying amount of time to complete

locriandev: This is a very good point, error handling must be improved

locriandev: > Tested it against RCs and its working heavy_check_mark When I tried with a release `@art-bot-dev what kernel is used in 44.10.26` I got this error though:
> 
> ```
> error: image "quay.io/openshift-release-dev/ocp-release:44.10.26-x86_64" not found: manifest unknown: manifest unknown
> ```
> 
> I'm not sure if its because I don't have enough authorization though.

I think you have mispelled the release nvr, trying with @art-bot-dev what kernel is used in 4.10.26 work fine. In the last commit (https://github.com/openshift/art-bot/pull/18/commits/4da2bbe3644d3e6a829513a3af77bd0261263cd8) I have handled this error, this is the new behaviour:
```
@art-bot-daniele what kernel is used in 44.10.26

Sorry, I wasn't able to query the release image 44.10.26.
Sorry, I encountered an error. Please contact @art-team with the following details.
[...] (Exception details)
```

ashwindasr: You're absolutely right. All three cases are working.

ashwindasr: I see! Got it.

artbotlib/buildinfo.py
```diff
@@ -122,3 +164,68 @@ def brew_build_url(nvr):
...
+            asyncio.ensure_future(rhcos_kernel_info())
+        ]
+        return await asyncio.gather(*tasks, return_exceptions=True)
+
+    # Format output and send to Slack
+    res = loop.run_until_complete(gather_kernel_info())
+    output = []
+    for entry in res:
+        if isinstance(entry, ChildProcessError):
+            so.say(f"Sorry, I wasn't able to query the release image {release_img}.")
```
ashwindasr, line 222: Can you output the error logs to the art-bot monitoring channel as well using `so.monitoring_say`? That way we can occasionally monitor and find if something is going wrong.

ashwindasr (approved):

PR #22, opened 2022-07-11
Update krb5-redhat.conf

ashwindasr:


ashwindasr:
We changed from `true` to `fallback` because[ Pyxis documentation](https://docs.engineering.redhat.com/display/HSSP/Pyxis+access+request) mentioned that. I'm not sure that their API will work after roll back. We could rebuild and see though.

sosiouxme:
while we're waiting on that... let other code touching brew work.

PR #23, opened 2022-07-14
Slack Notification for Nightly status change

ashwindasr:
Art-Bot will notify users if a blue nightly changes to green/red.
Usage:
```
@art-bot Alert if https://amd64.ocp.releases.ci.openshift.org/releasestream/4.10.0-0.nightly/release/4.10.0-0.nightly-2022-07-13-131411 stops being blue
ART-Bot: Ok, I'll respond here when tests have finished.
<after a long time>
@ashwin Color of amd64 nightly 4.10.0-0.nightly-2022-07-13-131411 changed to Green!
```
It will keep checking every 5 mins and time out after 12 hrs with a error message.
If the color is already green/red, it will report it and exit

### DIFF
--- a/art-bot.py
+++ b/art-bot.py
@@ -49,6 +49,7 @@ _*ART releases:*_
 * Which build of `image_name` is in `release image name or pullspec`?
 * What (commits|catalogs|distgits|nvrs|images) are associated with `release-tag`?
 * Image list advisory `advisory_id`
+* Alert if `release_url` (stops being blue|fails|is rejected|is red|is accepted|is green)
 
 _*ART build info:*_
 * Where in `major.minor` (is|are) the `name1,name2,...` (RPM|package) used?
@@ -280,7 +281,7 @@ def respond(client: RTMClient, event: dict):
 
             # Others
             {
-                'regex': r'^Alert if https://(?P<release_browser>[\w]+).ocp.releases.ci.openshift.org(?P<nightly_url>[\w/.-]+) stops being blue$',
+                'regex': r'^Alert if https://(?P<release_browser>[\w]+).ocp.releases.ci.openshift.org(?P<release_url>[\w/.-]+) (stops being blue|fails|is rejected|is red|is accepted|is green)$',
                 'flag': re.I,
                 'function': nightly_color_status,
                 'user_id': True
--- a/art-bot.py
+++ b/art-bot.py
@@ -22,6 +22,7 @@ from artbotlib.slack_output import SlackOutput
 from artbotlib import brew_list, elliott
 from artbotlib.pipeline_image_names import pipeline_from_distgit, pipeline_from_github, pipeline_from_brew, \
     pipeline_from_cdn, pipeline_from_delivery
+from artbotlib.nightly_color import nightly_color_status
 
 logger = logging.getLogger()
 
@@ -275,14 +276,24 @@ def respond(client: RTMClient, event: dict):
                 'regex': r'^.*(image )?pipeline \s*for \s*image \s*(registry.redhat.io/)*(openshift4/)*(?P<delivery_repo_name>[a-zA-Z0-9-]+)\s*( \s*in \s*(?P<version>\d+.\d+))?\s*$',
                 'flag': re.I,
                 'function': pipeline_from_delivery
-            }
+            },
 
+            # Others
+            {
+                'regex': r'^Alert if https://(?P<release_browser>[\w]+).ocp.releases.ci.openshift.org(?P<nightly_url>[\w/.-]+) stops being blue$',
+                'flag': re.I,
+                'function': nightly_color_status,
+                'user_id': True
+            }
         ]
 
         for r in regex_maps:
             m = re.match(r['regex'], plain_text, r['flag'])
             if m:
-                r['function'](so, **m.groupdict())
+                if r.get('user_id', False):  # if functions need to cc the user
+                    r['function'](so, user_id, **m.groupdict())
+                else:
+                    r['function'](so, **m.groupdict())
 
         if not so.said_something:
             so.say("Sorry, I can't help with that yet. Ask 'help' to see what I can do.")
@@ -353,7 +364,7 @@ def consumer_thread(client_id, topic, callback_handler, consumer, durable, user_
 
 @click.command()
 def run():
-
+    
     try:
         config_file = os.environ.get("ART_BOT_SETTINGS_YAML", f"{os.environ['HOME']}/.config/art-bot/settings.yaml")
         with open(config_file, 'r') as stream:
--- a/art-bot.py
+++ b/art-bot.py
@@ -281,7 +281,7 @@ def respond(client: RTMClient, event: dict):
 
             # Others
             {
-                'regex': r'^Alert if https://(?P<release_browser>[\w]+).ocp.releases.ci.openshift.org(?P<release_url>[\w/.-]+) (stops being blue|fails|is rejected|is red|is accepted|is green)$',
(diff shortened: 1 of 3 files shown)


artbotlib/nightly_color.py
```diff
@@ -0,0 +1,62 @@
...
+    Function to get the color of the nightly for the given release browser
+    :param nightly_url: URL of the nightly eg: /releasestream/4.10.0-0.nightly/release/4.10.0-0.nightly-2022-07-13-131411
+    :param release_browser: release browser eg: ppc64le/arm64/s390x/amd64
+    :return:
+    """
+    url = f"https://{release_browser}.ocp.releases.ci.openshift.org"
+    response = requests.get(url)  # get the webpage
+
+    data = response.content.decode()
+    soup = BeautifulSoup(data, "html.parser")  # parse html
```
Ximinhan, line 23: use api could be more accurate, check this https://github.com/Ximinhan/aos-cd-jobs/blob/master/pyartcd/pyartcd/pipelines/promote.py#L627

ashwindasr: Are the different states `Ready` for Blue, `Accepted` for Green and `Rejected` for Red?

ashwindasr: Updated!

artbotlib/nightly_color.py
```diff
@@ -0,0 +1,55 @@
...
+        color = get_nightly_color(nightly_url, release_browser)
+        if color:  # if color is not blue return the current color and exit
+            so.say(f"<@{user_id}> Color of nightly is already `{color}`")
+            return
+
+        so.say(f"<@{user_id}> Ok, I'll respond here when tests have finished.")
+        start = time.time()
+        while True:
+            now = time.time()
+            if now - start > 43200:  # Timeout after 12 hrs.
```
locriandev, line 44: I'd suggest a more explicit constant definition such as 

TWELWE_HOURS = 60 * 60 * 12

art-bot.py
```diff
@@ -275,14 +276,24 @@ def respond(client: RTMClient, event: dict):
                 'regex': r'^.*(image )?pipeline \s*for \s*image \s*(registry.redhat.io/)*(openshift4/)*(?P<delivery_repo_name>[a-zA-Z0-9-]+)\s*( \s*in \s*(?P<version>\d+.\d+))?\s*$',
                 'flag': re.I,
                 'function': pipeline_from_delivery
-            }
+            },
 
+            # Others
+            {
+                'regex': r'^Alert if https://(?P<release_browser>[\w]+).ocp.releases.ci.openshift.org(?P<nightly_url>[\w/.-]+) stops being blue$',
```
thegreyd, line 283: This is good!
I'd suggest having the nightly additional prompts as - 
- Alert if <nightly/release> fails/is rejected
- Alert if <nightly/release> is accepted

artbotlib/nightly_color.py
```diff
@@ -0,0 +1,58 @@
...
+            now = time.time()
+            if now - start > TWELVE_HOURS:  # Timeout after 12 hrs.
+                so.say(f"<@{user_id}> Color didn't change even after 12 hrs :(")
+                break
+
+            time.sleep(FIVE_MINUTES)  # check every 5 minutes
+
+            color = get_nightly_color(nightly_url, release_browser)
+            if color:
+                so.say(f"<@{user_id}> Color of {release_browser} nightly {nightly_url.split('/')[-1]} changed to `{color}`!")
```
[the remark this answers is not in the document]
thegreyd: Example https://amd64.ocp.releases.ci.openshift.org/api/v1/releasestream/4-stable/release/4.11.0-fc.2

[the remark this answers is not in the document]
ashwindasr: Okay!

art-bot.py
locriandev: not sure if this is going to add much value, we usually wait until blocking jobs are done and hope it gets accepted, so the most frequent use case I see is "art bot let me know when the blocking jobs are done so I can decide what to do". But maybe there's other scenarios I am not thinking of

artbotlib/nightly_color.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def get_nightly_color(nightly_url, release_browser) -> Union[str, None]:
+    """
+    Function to get the color of the nightly for the given release browser
+    :param nightly_url: URL of the nightly eg: /releasestream/4.10.0-0.nightly/release/4.10.0-0.nightly-2022-07-13-131411
+    :param release_browser: release browser eg: ppc64le/arm64/s390x/amd64
+    :return:
+    """
+    url = f"https://{release_browser}.ocp.releases.ci.openshift.org/api/v1{nightly_url}"
```
thegreyd, line 21: This url I think would also work for releases, so if you pass in 
https://amd64.ocp.releases.ci.openshift.org/releasestream/4-stable/release/4.10.19
So we can replace the word "nightly" with something generic, since we'll also use this for releases.

artbotlib/nightly_color.py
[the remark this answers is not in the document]
thegreyd: Ah forgot about nightly. So when a nightly is red, then we'll look at the blockingJobs portion of the response and pass in those that failed.

artbotlib/nightly_color.py
ashwindasr: Got it! Will work on that!

art-bot.py
ashwindasr: Hmm... well all the cases will call the same function only and will notify if a color changes to Red or Green (or timesout)

thegreyd: Yeah so when an rc blocking job fails (which will turn the release red), we want to 1) trigger more tests (if not already), 2) wait for a test to pass, and then 3) paint the release green and then 4) re-run the promote job. Right now 1,2,3,4 are all manual. We can use those for the different stages we're at, also since it's just 2 more regex's to add, don't think the effort is much right?

ashwindasr: Updated regex!

artbotlib/nightly_color.py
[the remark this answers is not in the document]
ashwindasr: Updated!

artbotlib/nightly_color.py
ashwindasr: Updated!

sosiouxme (approved):
/lgtm

art-bot.py
```diff
@@ -275,14 +277,24 @@ def respond(client: RTMClient, event: dict):
                 'regex': r'^.*(image )?pipeline \s*for \s*image \s*(registry.redhat.io/)*(openshift4/)*(?P<delivery_repo_name>[a-zA-Z0-9-]+)\s*( \s*in \s*(?P<version>\d+.\d+))?\s*$',
                 'flag': re.I,
                 'function': pipeline_from_delivery
-            }
+            },
 
+            # Others
+            {
+                'regex': r'^Alert if https://(?P<release_browser>[\w]+).ocp.releases.ci.openshift.org(?P<release_url>[\w/.-]+) (stops being blue|fails|is rejected|is red|is accepted|is green)$',
```
sosiouxme, line 284: I'd probably make this so I don't have to remember:
```suggestion
                'regex': r'^Alert ?(if|when|on)? https://(?P<release_browser>[\w]+).ocp.releases.ci.openshift.org(?P<release_url>[\w/.-]+) ?(stops being blue|fails|is rejected|is red|is accepted|is green)?$',
```
:shrug:

ashwindasr: Done!

PR #26, opened 2022-07-30
kerberos issue fix

ashwindasr:
Art-bot kerberos had an issue in which the keytab could not loaded. Fixed the issue by only reading the credentials from the mounted keytab and doing `kinit`. The new kerberos credentials will be stored at the default location which kerberos will automatically access. Reused functions from art-dash codebase.

### DIFF
--- a/artbotlib/kerberos.py
+++ b/artbotlib/kerberos.py
@@ -1,0 +1,37 @@
+import os
+import time
+import subprocess
+
+
+def update_last_kinit_env_var():
+    """
+    This function performs kinit as and when required.
+    The function expects .aws directory in the project base directory.
+    The .aws directory should have a keytab file by name redhat.keytab.
+    :return: None
+    """
+
+    keytab_file = "/tmp/keytab/keytab"
+    kinit_request = subprocess.Popen(["kinit", "-kt", keytab_file, "ocp-build/buildvm.openshift.eng.bos.redhat.com@IPA.REDHAT.COM"],
+                                     stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    output, error = kinit_request.communicate()
+    if error:
+        print(error)
+    os.environ["LAST_KINIT_TIME"] = str(int(time.time()))
+
+
+def handle_kinit():
+    """
+    This function tests whether kinit is required. The current policy is to do kinit after
+    every hour.
+    :return: None
+    """
+
+    if "LAST_KINIT_TIME" not in os.environ:
+        update_last_kinit_env_var()
+
+    last_update = os.environ["LAST_KINIT_TIME"]
+    time_now = int(time.time())
+
+    if time_now - int(last_update) >= 3600:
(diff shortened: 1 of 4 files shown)


ashwindasr:
Also need to remove `KRB5CCNAME` from build config so that it doesn't get set as an env variable.

artbotlib/kerberos.py
```diff
@@ -0,0 +1,37 @@
...
+    :return: None
+    """
+
+    keytab_file = "/tmp/keytab/keytab"
+    kinit_request = subprocess.Popen(["kinit", "-kt", keytab_file, "ocp-build/buildvm.openshift.eng.bos.redhat.com@IPA.REDHAT.COM"],
+                                     stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    output, error = kinit_request.communicate()
+    if error:
+        print(error)
+    os.environ["LAST_KINIT_TIME"] = str(int(time.time()))
```
sosiouxme, line 20: why an env var? just seems unnecessary, you could just set a local var like so:
```suggestion

last_kinit_timestamp: str = ""


def update_last_kinit_var():
    """
    This function performs kinit as and when required.
    The function expects .aws directory in the project base directory.
    The .aws directory should have a keytab file by name redhat.keytab.
    :return: None
    """

    keytab_file = "/tmp/keytab/keytab"
    kinit_request = subprocess.Popen(["kinit", "-kt", keytab_file, "ocp-build/buildvm.openshift.eng.bos.redhat.com@IPA.REDHAT.COM"],
                                     stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    output, error = kinit_request.communicate()
    if error:
        print(error)
    last_kinit_timestamp = str(int(time.time()))
```

sosiouxme (commented):
I'm not sure we really need to limit kinit at all, pretty sure buildvm does it hundreds of times an hour. But since we already have the code, I guess it's nice to rate limit ourselves... just, if it causes any problems, it's ok to rip it out and kinit every time.

ashwindasr:
Understood. I reused the code from art-dash since it was working. Will test out these changes as well.

artbotlib/util.py
```diff
@@ -208,3 +210,12 @@ def cached_ttl(func):
     def wrapper(*args, **kwargs):
         return func(*args, **kwargs)
     return wrapper
+
+
+def update_keytab(func):
```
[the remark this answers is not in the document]
ashwindasr: Done!

artbotlib/kerberos.py
ashwindasr: I changed to always kinit

sosiouxme (approved):
i'd like to make this configurable so i can specify the location of the keytab in config, and if it's not configured then it just uses `klist -s` to check if there's already a ticket. but i probably have a thing or two to tweak to get my setup working, no need to hold this up. thanks!

PR #37, opened 2022-11-04
Add developer interface for art-bot

ashwindasr:
Preparing art-bot for migrating from classic slack app to the latest slack app. Currently slack is sending responses to both production and development environments as its using [RTM](https://api.slack.com/rtm). When we migrate to socket mode, slack will randomly send the responses to [any of the connections](https://api.slack.com/apis/connections/socket-implement#connections). Therefore developing using `@art-bot-dev` will no longer be possible.

It was proposed during team sync that a developer command line interface could remedy the situation, as we can test the regexes using that and in production, we'll connect to the actual slack object.

To start, run `./art_bot_dev`

Sample interface:
```
Welcome to the developer interface for Art-Bot.

Enter your command ('exit' to break): what's up
'what's up' did not match any regexes.

Enter your command ('exit' to break): hello
so.say:
---
Hi, @developer
---
Enter your command ('exit' to break): exit
Exiting...
```

Need to set NEEDS_KINIT in production after merge.

### DIFF
--- a/art_bot_dev.py
+++ b/art_bot_dev.py
@@ -4,9 +4,9 @@
 
 from artbotlib.slack_output import SlackDeveloperOutput
 
-art_bot = __import__('art-bot')
+art_bot = __import__("art-bot")
 
-if __name__ == '__main__':
+if __name__ == "__main__":
     so = SlackDeveloperOutput()
     print("---\nWelcome to the developer interface for Art-Bot.")
     print("To exit, type in 'exit' or use Ctrl-C\n---\n")
@@ -17,4 +17,4 @@ if __name__ == '__main__':
                 break
             art_bot.map_command_to_regex(so, command, None)
     except KeyboardInterrupt:
-        print('Exiting...')
+        print("Exiting...")
--- a/art_bot_dev.py
+++ b/art_bot_dev.py
@@ -1,0 +1,20 @@
+#!/usr/bin/python3
+
+# Script to start the developer interface for art-bot
+
+from artbotlib.slack_output import SlackDeveloperOutput
+
+art_bot = __import__('art-bot')
+
+if __name__ == '__main__':
+    so = SlackDeveloperOutput()
+    print("---\nWelcome to the developer interface for Art-Bot.")
+    print("To exit, type in 'exit' or use Ctrl-C\n---\n")
+    try:
+        while True:
+            command = input("Enter your command: ")
+            if command.lower() == "exit":
+                break
+            art_bot.map_command_to_regex(so, command, None)
+    except KeyboardInterrupt:
+        print('Exiting...')
--- a/artbotlib/slack_output.py
+++ b/artbotlib/slack_output.py
@@ -15,7 +15,7 @@ class SlackOutput:
         self.said_something = False
 
     def say(self, text, **msg_opts):
-        if os.environ.get("RUN_ENV") == 'production':
+        if os.environ.get("RUN_ENV") == "production":
             print(f"Responding back through: {self.target_channel_id}")
             self.said_something = True
             msg = dict(
@@ -50,7 +50,7 @@ class SlackOutput:
         pprint.pprint(r)
 
     def monitoring_say(self, text, **msg_opts):
-        if os.environ.get("RUN_ENV") == 'production':
+        if os.environ.get("RUN_ENV") == "production":
             if not self.monitoring_channel_id:
                 return
             try:
@@ -89,7 +89,7 @@ class SlackOutput:
             traceback.print_exc()
 
     def from_user_mention(self):
-        if os.environ.get("RUN_ENV") == 'production':
+        if os.environ.get("RUN_ENV") == "production":
             return f"<@{self.from_user_id()}>"
         return "@developer"
 
--- a/artbotlib/slack_output.py
+++ b/artbotlib/slack_output.py
@@ -1,5 +1,6 @@
 import pprint
 import traceback
+import os
 
 
 class SlackOutput:
@@ -14,18 +15,24 @@ class SlackOutput:
         self.said_something = False
 
     def say(self, text, **msg_opts):
-        print(f"Responding back through: {self.target_channel_id}")
-        self.said_something = True
-        msg = dict(
-            channel=self.target_channel_id,
-            text=text,
-            thread_ts=self.thread_ts,
-            **self.username_opts,
-        )
-        msg.update(msg_opts)
-        response = self.web_client.chat_postMessage(**msg)
-        print(f"response ok: {response.get('ok')}\n")
-        pprint.pprint(f"ok: {response.get('message')}")
+        if os.environ.get("RUN_ENV") == 'production':
+            print(f"Responding back through: {self.target_channel_id}")
+            self.said_something = True
+            msg = dict(
+                channel=self.target_channel_id,
+                text=text,
+                thread_ts=self.thread_ts,
+                **self.username_opts,
+            )
+            msg.update(msg_opts)
+            response = self.web_client.chat_postMessage(**msg)
+            print(f"response ok: {response.get('ok')}\n")
+            pprint.pprint(f"ok: {response.get('message')}")
+        else:
+            print("so.say:")
+            print("---")
+            print(text)
+            print("---")
 
     def snippet(self, payload, intro=None, filename=None, filetype=None):
         self.said_something = True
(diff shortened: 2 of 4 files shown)


locriandev (commented):
I've tested a query like `what images build in 4.11`, the CLI crashed by `AttributeError`.

locriandev:
By wrapping the `command = input("Enter your command ('exit' to break): ")` with a `try/except` block we will let the user terminate the CLI by typing `CTRL^C`

locriandev:
I would rather not have to define and env var for production; maybe we can turn the logic around, by forcing the user to define `RUN_ENV=test` when needed?

locriandev:
I think it would be cleaner to have the CLI as a separate executable. We might move `regex_maps` outside the local scope (better if to a dedicated module) and create a `cli.py` that imports it

ashwindasr:
> I would rather not have to define and env var for production; maybe we can turn the logic around, by forcing the user to define `RUN_ENV=test` when needed?

Yes I first thought about that, but the issue is that if the developer forgets to add  `RUN_ENV=test` there will be two production instances. What about `RUN_ENV != production`, that should enforce the same constraint right?

ashwindasr:
> By wrapping the `command = input("Enter your command ('exit' to break): ")` with a `try/except` block we will let the user terminate the CLI by typing `CTRL^C`

I tried to do that, but it doesn't seem to be working, since we're waiting for an input `''` is going through.

ashwindasr:
> I've tested a query like `what images build in 4.11`, the CLI crashed by `AttributeError`.

It was because of the `SlackOutput` class. Instead of checking for production everytime, I created a new class which inherits from `SlackOutput` class and overrode the functions to just print.

locriandev:
> I tried to do that, but it doesn't seem to be working, since we're waiting for an input '' is going through.

CTRL^C works for me: https://github.com/ashwindasr/art-bot/compare/development-command-line-feature...locriandev:art-bot:development-command-line-feature

locriandev:
> > I would rather not have to define and env var for production; maybe we can turn the logic around, by forcing the user to define `RUN_ENV=test` when needed?
> 
> Yes I first thought about that, but the issue is that if the developer forgets to add `RUN_ENV=test` there will be two production instances and with the new slack bot, that can cause a lot of issues. We're actually doing this for art-dash to determine which env files to load.

I think this is another reason in favor of decoupling concerns. Our CLI applications does not need to know anything about Slack, it's just there to test out regex. Making it a separate executable would solve this issue

ashwindasr:
> CTRL^C works for me: [ashwindasr/art-bot@development-command-line-feature...locriandev:art-bot:development-command-line-feature](https://github.com/ashwindasr/art-bot/compare/development-command-line-feature...locriandev:art-bot:development-command-line-feature)

Ahh I see. I think its an issue with my setup. I was trying to remotely connect to the container using PyCharm for which Ctrl-C doesn't work, but when I run the script it seems to be working now. Will push the change thank you!

ashwindasr:
> > > I would rather not have to define and env var for production; maybe we can turn the logic around, by forcing the user to define `RUN_ENV=test` when needed?
> > 
> > 
> > Yes I first thought about that, but the issue is that if the developer forgets to add `RUN_ENV=test` there will be two production instances and with the new slack bot, that can cause a lot of issues. We're actually doing this for art-dash to determine which env files to load.
> 
> I think this is another reason in favor of decoupling concerns. Our CLI applications does not need to know anything about Slack, it's just there to test out regex. Making it a separate executable would solve this issue

Understood. Will create a new script.

art_bot_dev.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/python3
+
+# Script to start the developer interface for art-bot
+
+from artbotlib.slack_output import SlackDeveloperOutput
+
+art_bot = __import__("art-bot")
```
ashwindasr, line 7: Since dashes are present in `art-bot.py`, we can't import functions from that file by saying `import art-bot`. We either have to rename the file to `art_bot.py` by replacing the dashes with underscores or use the `__import__` function, as used here. The renaming will result in a change at a lot of places including the deployment in production, but if that's the better way to go we can do that as well.

ashwindasr:
Need to set `NEEDS_KINIT` in production after merge.

art_bot_dev.py
ashwindasr: Decided to refactor after slack app migration.

artbotlib/slack_output.py
```diff
@@ -27,6 +28,7 @@ def say(self, text, **msg_opts):
         print(f"response ok: {response.get('ok')}\n")
         pprint.pprint(f"ok: {response.get('message')}")
 
+
```
locriandev, line 31: Remove extra blank line (only one needed between class methods)

artbotlib/slack_output.py
```diff
@@ -53,10 +55,11 @@ def monitoring_say(self, text, **msg_opts):
             )
             msg.update(msg_opts)
             self.web_client.chat_postMessage(**msg)
-        except:
+        except Exception:
             print("Error sending information to monitoring channel")
             traceback.print_exc()
 
+
```
locriandev, line 62: Remove extra blank line (only one needed between class methods)

artbotlib/slack_output.py
```diff
@@ -83,3 +86,54 @@ def from_user_id(self):
 
     def from_channel(self):
         return self.event.get("channel", None)
+
+
+def print_payload(text):
+    print("---")
+    print(text)
+    print("---")
+
```
locriandev, line 95: Add another blank line (two needed between module-level functions)

artbotlib/slack_output.py
```diff
@@ -83,3 +86,54 @@ def from_user_id(self):
...
+    if filetype:
+        print("filetype:")
+        print(filetype)
+    print("---")
+
+
+class SlackDeveloperOutput(SlackOutput):
+    def __init__(self, web_client=None, event=None, target_channel_id=None, monitoring_channel_id=None, thread_ts=None, alt_username=None):
+        super().__init__(web_client, event, target_channel_id, monitoring_channel_id, thread_ts, alt_username)
+    def say(self, text, **msg_opts):
```
locriandev, line 115: Add a blank line before method definition

artbotlib/slack_output.py
```diff
@@ -83,3 +86,54 @@ def from_user_id(self):
...
+        print("so.say:")
+        print_payload(text)
+
+
```
locriandev, line 119: Remove extra blank line (only one needed between class methods)

artbotlib/slack_output.py
```diff
@@ -83,3 +86,54 @@ def from_user_id(self):
...
+        print_snippet_payload(payload, intro, filename, filetype)
+
+    def from_user_mention(self):
+        return "@developer"
+
+    def from_user_id(self):
+        return "@developer"
+
+    def from_channel(self):
+        return "developer_channel"
```
locriandev, line 139: Add a new line at the end of the file

locriandev (changes requested):

artbotlib/slack_output.py
ashwindasr: Fixed in https://github.com/openshift/art-bot/pull/37/commits/843bc5d60bb27653934f62f5f2c40b89bb426753

artbotlib/slack_output.py
ashwindasr: Fixed in https://github.com/openshift/art-bot/pull/37/commits/843bc5d60bb27653934f62f5f2c40b89bb426753

artbotlib/slack_output.py
ashwindasr: Fixed in https://github.com/openshift/art-bot/pull/37/commits/843bc5d60bb27653934f62f5f2c40b89bb426753

artbotlib/slack_output.py
ashwindasr: Fixed in https://github.com/openshift/art-bot/pull/37/commits/843bc5d60bb27653934f62f5f2c40b89bb426753

artbotlib/slack_output.py
ashwindasr: Fixed in https://github.com/openshift/art-bot/pull/37/commits/843bc5d60bb27653934f62f5f2c40b89bb426753

artbotlib/slack_output.py
ashwindasr: Fixed in https://github.com/openshift/art-bot/pull/37/commits/843bc5d60bb27653934f62f5f2c40b89bb426753

PR #44, opened 2022-11-29
Add GitHub actions

ashwindasr:
It was discussed in the arch-call to test out both GitHub Actions and Prow before deciding to move forward.

Config: https://github.com/ashwindasr/art-bot/blob/github-actions/.github/workflows/run_tox.yml
Sample Actions run: https://github.com/ashwindasr/art-bot/actions/runs/3575852343

The tests are failing because of flake8 errors.


tox.ini
```diff
@@ -11,16 +15,32 @@ commands =
...
+# variable in function should be lowercase
+    N806,
+# redefinition of unused
+    F811
+# lambda
+    E731,
+# bare except
+    E722,
+# line break before binary operator
+    W503,
```
sosiouxme, line 43: these mostly look like legitimate complaints... why not fix the source? or make the exception at the offending source line?

ashwindasr: Yeah! I am working on the fixes locally. Reused the tox file from `elliott` which had these exceptions.

ashwindasr: Fixed in https://github.com/openshift-eng/art-bot/pull/44/commits/daaadd794270f935169dbc4e24e47c276856b124

ashwindasr:
Updated `flexmock` to `flexmock.flexmock` to fix module not found error.
All tests passing: https://github.com/ashwindasr/art-bot/actions/runs/3594040595

tox.ini
```diff
@@ -11,16 +15,18 @@ commands =
...
     F401,
-    # Line too long
+# Line too long
     E501,
-    # multiple imports on one line
-    E401,
-    # lambda
-    E731
+# Indentation
+    E126,
```
joepvd, line 25: We do check for this in eliott and doozer code. My vote would be to keep on checking on this.

joepvd (approved):
/lgtm

With an optional nit.
Maybe you could `git rebase -i origin/master` so the commit history is a bit cleaner.

tox.ini
ashwindasr: Got it. Fixed in https://github.com/openshift-eng/art-bot/pull/44/commits/a7b515feeddf10c3dce00f09142d6f261f495778

sosiouxme (approved):
would like "real" changes in separate commits from flake8 fixes, but lgtm

PR #46, opened 2022-12-02
Catch unexpected errors

ashwindasr:
There was a `missing_scope` error that was not caught and caused no error messages to be displayed to the users invoking the command. Hence adding a try-except for `map_command_to_regex` function for now to catch any other unexpected errors so that the user can be notified and art-team can resolve the issue quickly since it'll be logged in the monitoring channel.


sosiouxme (approved):

ashwindasr:
Rebuilding and redeploying.

PR #47, opened 2022-12-03
Report to user if pod restarted

ashwindasr:
The `alert nightly/release` command will constantly check the release status page and once the color changes to green or red, it'll report back to the user. But if the pod is restarted due to some reason, while the command is still running, the user will not be notified. Hence by capturing  the `SIGTERM` signal that Kubernetes sends before terminating ([docs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination)), we can send a message back to the user saying that the pod is restarted and the command has to be run again.


art-bot.py
```diff
@@ -378,6 +374,17 @@ def run(debug):
     handler = SocketModeHandler(app, bot_config["slack_app_token"])
     handler.start()
 
+def signal_handler(_signo, _stack_frame):
+    """
+    Function to handle SIGTERM signal sent during pod termination. Check if check nightly color command is still running.
+    If it is, report to the user saying that they have re-run the command.
+    """
+    if nightly_color.ERROR_SLACK_OBJECT:
+        nightly_color.ERROR_SLACK_OBJECT.say("Uh oh... the pod was restarted. Please run the command again.")
```
joepvd, line 383: Maybe quote what was asked, if that is still available? Maybe in a long thread it is not clear what this is about.

art-bot.py
```diff
@@ -325,12 +326,7 @@ def respond(client, event):
 
         so.monitoring_say(f"<@{user_id}> asked: {plain_text}")
 
-        try:
-            map_command_to_regex(so, plain_text, user_id)
-        except Exception as error:
-            # Catch any unexpected error and display appropriate message to the user.
-            so.say("Uh oh... there seems to be a problem. Please contact @.art-team")
-            so.monitoring_say(f"Error: {error}")
+        map_command_to_regex(so, plain_text, user_id)
```
joepvd, line 329: Any reason this exception disappeared? Think it should stay. Or does it interfere with the signal work?

art-bot.py
```diff
@@ -378,6 +374,17 @@ def run(debug):
...
     handler.start()
 
+def signal_handler(_signo, _stack_frame):
+    """
+    Function to handle SIGTERM signal sent during pod termination. Check if check nightly color command is still running.
+    If it is, report to the user saying that they have re-run the command.
+    """
+    if nightly_color.ERROR_SLACK_OBJECT:
+        nightly_color.ERROR_SLACK_OBJECT.say("Uh oh... the pod was restarted. Please run the command again.")
+    sys.exit(0)
```
joepvd, line 384: What happens if there are two requests for updates on payload status? Think without `exit`, it'd work just as fine. Have a close look at threading here :)

joepvd (commented):
Left some remarks

art-bot.py
ashwindasr: Ahh my bad, I was working on those two features and forgot to update the branch. Fixed! https://github.com/openshift-eng/art-bot/pull/47/commits/a0aef491f65e18190cf83ef8d4516d8eb655a65f

art-bot.py
ashwindasr: Removed `exit`. Yeah I need to start looking at threading seriously. Thank you!

ashwindasr:
Added support to handle multiple threads

art-bot.py
```diff
@@ -379,5 +379,19 @@ def run(debug):
...
+def signal_handler(_signo, _stack_frame):
+    """
+    Function to handle SIGTERM signal sent during pod termination. Check if check nightly color command is still running.
+    If it is, report to the user saying that they have re-run the command.
+    """
+    slack_objects = variables.active_slack_objects
+    if slack_objects:
+        for slack_object in slack_objects:
+            plain_text = extract_plain_text({"data": slack_object.event})
+            slack_object.say(f"Uh oh... the pod was restarted. Please run *'{plain_text}'* again.")
```
sosiouxme, line 391: "the pod" seems like a non sequitur here if you know nothing about art-bot infra - how about
```suggestion
            slack_object.say(f"Uh oh... art-bot is restarting. Please try *'{plain_text}'* again.")
```

artbotlib/nightly_color.py
```diff
@@ -71,10 +71,12 @@ def nightly_color_status(so, user_id, release_url, release_browser) -> None:
 
     so.say(f"<@{user_id}> Ok, I'll respond here when tests have finished.")
     start = time.time()
+    variables.active_slack_objects[so] = 1
```
sosiouxme, line 74: i'm not sure what happens if we leak `so` instances but i'd rather not find out. how about wrapping everything that follows in a `try` and using the `finally` to `del variables.active_slack_objects[so]`?

also why is this a dict and not a set?

sosiouxme (commented):
few things to change, but on the whole seems good!

artbotlib/nightly_color.py
ashwindasr: Got it will use a `try-finally`.

Well I was thinking to remove the object in constant time...

art-bot.py
ashwindasr: Fixed in https://github.com/openshift-eng/art-bot/pull/47/commits/d6684750d95a85bee51ee9ec564d28f9e58ca02a

art-bot.py
```diff
@@ -379,5 +379,19 @@ def run(debug):
...
 
 
+def signal_handler(_signo, _stack_frame):
+    """
+    Function to handle SIGTERM signal sent during pod termination. Check if check nightly color command is still running.
+    If it is, report to the user saying that they have re-run the command.
+    """
+    slack_objects = variables.active_slack_objects
+    if slack_objects:
+        for slack_object in slack_objects:
```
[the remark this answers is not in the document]
ashwindasr: Fixed in https://github.com/openshift-eng/art-bot/pull/47/commits/d6684750d95a85bee51ee9ec564d28f9e58ca02a

PR #50, opened 2022-12-15
Move exec functions in a dedicated module

locriandev:


artbotlib/brew_list.py
```diff
@@ -364,7 +365,7 @@ def _tags_for_version(major_minor):
 
 def list_images_in_major_minor(so, major, minor):
     major_minor = f'{major}.{minor}'
-    rc, stdout, stderr = util.cmd_assert(so,
+    rc, stdout, stderr = artbotlib.exectools.cmd_assert(so,
```
ashwindasr, line 368: This as well

artbotlib/brew_list.py
```diff
@@ -148,7 +149,7 @@ def latest_images_for_version(so, major_minor):
     so.say(f"Determining images for {major_minor} - this may take a few minutes...")
 
     try:
-        rc, stdout, stderr = util.cmd_assert(so,
+        rc, stdout, stderr = artbotlib.exectools.cmd_assert(so,
```
ashwindasr, line 152: To fix this flake8 error this has to be in a single line.

ashwindasr (changes requested):
Might wanna update path `tox.ini ` line 27 from `artbotlib/util.py:F821` to `artbotlib/exectools.py:F821` . If you can fix without suppressing, that would be great as well! I tried, but couldn't find the solution

ashwindasr (approved):

PR #54, opened 2022-12-18
Update kerberos setup, README

ashwindasr:
Updates in this PR:
- Added art-dash-server route as a constant to the repo.
- Kerberos checks `/etc/krb5.conf` first before `/etc/krb5.conf.d/*`. So its best to keep the file in the default location, to prevent confusion. 
- Since art-bot development has been decoupled from slack, @.art-bot-dev is not longer used in slack. Hence there's no need to copy the `settings.yaml` in development anymore. UMB has also been removed since its not used. 
- slack-app-token field added to `settings.yaml` as well.


ashwindasr:
Mounting to `/etc/krb5.conf` did not work in production dev. Keeping the config as is for now. Will create a new PR to update docs.

artbotlib/constants.py
```diff
@@ -22,3 +22,5 @@
 ERRATA_TOOL_URL = 'https://errata.devel.redhat.com'
 
 GITHUB_API_OPENSHIFT = "https://api.github.com/repos/openshift"
+
+ART_DASH_SERVER_ROUTE = "http://art-dash-server-art-build-dev.apps.ocp4.prod.psi.redhat.com"
```
locriandev, line 26:
```suggestion
ART_DASH_SERVER_ROUTE = "http://art-dash-server-art-build-dev.apps.ocp4.prod.psi.redhat.com/api/v1"
```

locriandev (changes requested):

locriandev:
Generally lgtm, apart from the definition of `ART_DASH_SERVER_ROUTE` that should include `/api/v1` unless we think that the server base URL will be useful somewhere else.

What is `slack_app_token_file` needed for? How should we set that in production? (haven't found this anywhere)

ashwindasr:
> What is slack_app_token_file needed for? How should we set that in production? (haven't found this anywhere)


Its defined here: https://console-openshift-console.apps.ocp4.prod.psi.redhat.com/k8s/ns/aos-art-web/secrets/art-bot-config. The `art-bot.py` has code that pulls it from the env.

ashwindasr:
> that should include /api/v1 unless we think that the server base URL will be useful somewhere else.

You're right. Will update.

artbotlib/constants.py
ashwindasr: Fixed!

locriandev (approved):

PR #55, opened 2022-12-19
Add unit tests for regex mapping

locriandev:
Regex mapping unit tests can be used to check valid/invalid queries for all art-bot functionalities. They can also be forwarded to potential users in need for help.

Ref. https://issues.redhat.com/browse/ART-5340

### DIFF
--- a/artbotlib/pr_in_build.py
+++ b/artbotlib/pr_in_build.py
@@ -13,7 +13,6 @@ from artbotlib import util, pipeline_image_util
 from artbotlib.exceptions import NullDataReturned
 from artbotlib.constants import BREW_TASK_STATES, BREW_URL, GITHUB_API_OPENSHIFT
 
-API = f"{os.environ['ART_DASH_SERVER_ROUTE']}/api/v1"
 RELEASESTREAM_ENDPOINT_TEMPLATE = Template('https://${arch}.ocp.releases.ci.openshift.org/api/v1/releasestream')
 VALID_ARCHES = [
     'amd64',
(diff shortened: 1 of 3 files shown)


artbotlib/pr_in_build.py
```diff
@@ -13,7 +13,6 @@
 from artbotlib.exceptions import NullDataReturned
 from artbotlib.constants import BREW_TASK_STATES, BREW_URL, GITHUB_API_OPENSHIFT
 
-API = f"{os.environ['ART_DASH_SERVER_ROUTE']}/api/v1"
```
locriandev, line 16: This change is needed because when the module is imported, the importer will try to resolve ART_DASH_SERVER_ROUTE (undefined when unit testing). Moving it within the function body where it's actually used will prevent this from happening.

artbotlib/translation.py
```diff
@@ -2,14 +2,7 @@
 from . import util
 
 
-def translate_names(so, name_type, name, name_type2, major=None, minor=None):
```
locriandev, line 5: This lines can be removed as the check on valid types is already performed by the regular expression

artbotlib/pr_in_build.py
ashwindasr: I have a similar change in https://github.com/openshift-eng/art-bot/pull/54, will redefining it as a constant work?

locriandev: If it's not an env var, it will work no matter where it's placed. I'd suggest to define something like
```ART_DASH_API = 'http://art-dash-server-art-build-dev.apps.ocp4.prod.psi.redhat.com/api/v1'```,
since it's the only endpoint that matters to us

ashwindasr: Okay will do

ashwindasr (approved):

PR #57, opened 2023-01-03
Add art_bot_dev.py to production image

ashwindasr:
There are some cases in which the commands are working correctly in local development but doesn't seem to work in production. By adding the file, we can test out the commands using the terminal on Openshift.


ashwindasr:
Testing in `art build dev` namespace before merge.

PR #59, opened 2023-01-10
Poll for a Prow job state

locriandev:
Ref. https://issues.redhat.com/browse/ART-5685

### DIFF
--- a/artbotlib/prow.py
+++ b/artbotlib/prow.py
@@ -1,0 +1,158 @@
+import asyncio
+import json
+import logging
+import time
+from enum import Enum
+
+import aiohttp
+from aiohttp import client_exceptions
+
+from artbotlib.constants import FIVE_MINUTES, TWELVE_HOURS, PROW_BASE_URL
+
+GS_WEB_URL = 'https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs'
+
+logger = logging.getLogger(__name__)
+
+
+class ProwJobState(Enum):
+    PENDING = 'pending'
+    FAILURE = 'failure'
+    SUCCESS = 'success'
+
+
+async def get_job_state(job_path: str) -> str:
+    """
+    Gets Prow job state. Raises:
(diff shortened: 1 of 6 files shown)


artbotlib/prow.py
```diff
@@ -0,0 +1,158 @@
...
+    # This counter accounts for temporary outages (all jobs would fail)
+    fail_count = 0
+    max_attempts = 3
+
+    while True:
+        # Timeout after 12 hrs.
+        now = time.time()
+        if now - start > TWELVE_HOURS:  # Timeout after 12 hrs.
+            so.say(f"<@{user_id}> no job in {job_paths} completed within 12 hrs :(")
+            return
```
ashwindasr, line 120: It might be a good idea to ping to the user if the pod is restarted while this loop is running right? Like for nightly/release color.

ashwindasr (changes requested):

artbotlib/regex_mapping.py
```diff
@@ -141,6 +143,18 @@ def map_command_to_regex(so, plain_text, user_id):
...
             "user_id": True
+        },
+        {
+            "regex": rf"^Alert ?(if|when|on)? prow job {PROW_BASE_URL}/view/gs/(?P<job_path>\S*) completes$",
+            "flag": re.I,
+            "function": prow_job_status,
+            "user_id": True
+        },
+        {
+            "regex": rf"^Alert ?(if|when|on)? first prow job in(?P<job_paths>(( )*{PROW_BASE_URL}/view/gs/\S*)*) succeedes$",
```
ashwindasr, line 154: Also a typo `succeeds`, should be `succeeds` right. Here and other places as well

locriandev: right, fixed

artbotlib/prow.py
locriandev: Done in https://github.com/openshift-eng/art-bot/pull/59/commits/57757ebe5e7b207a1894514aebfe5ed771fb9eb1

PR #67, opened 2023-01-31
Refactor image pipeline code to use API endpoint

ashwindasr:
The image pipeline code was migrated to the back-end API server (endpoint: https://github.com/openshift-eng/art-dashboard-server/tree/master/api#get-apiv1pipeline-image). Refactor the code to use the data from the API server. Due to some complications with the regex, keeping multiple functions for each starting point eg: `github`, `cdn` etc. Will make the change in a separate PR.


artbotlib/pipeline_image_names.py
```diff
@@ -1,5 +1,64 @@
...
+    distgits = response_data['distgit']
+
+    for distgit in distgits:
+        payload += f"Production dist-git repo: " \
+                   f"<{distgit['distgit_url']}|*{distgit['distgit_repo_name']}*>\n"
+        payload += f"Production brew builds: " \
+                   f"<{distgit['brew']['brew_build_url']}|" \
+                   f"*{distgit['brew']['brew_package_name']}*>\n"
+
+        if distgit['brew']['bundle_component'] != "None":
```
locriandev, line 24: Wouldn't it be better to have art-dash response with `null` instead?
Then we could check it simply by `if distgit['brew']['bundle_component']:`

artbotlib/pipeline_image_names.py
```diff
@@ -1,5 +1,64 @@
...
+            payload += f"Bundle Component: *{distgit['brew']['bundle_component']}*\n"
+
+        if distgit['brew']['bundle_distgit'] != "None":
+            payload += f"Bundle Component: *{distgit['brew']['bundle_distgit']}*\n"
+
+        cdns = distgit['brew']['cdn']
+        if len(cdns) > 1:
```
locriandev, line 31:
```suggestion
        if cdns:
```

artbotlib/pipeline_image_names.py
```diff
@@ -1,5 +1,64 @@
-from artbotlib import exceptions, constants
-from artbotlib import pipeline_image_util
+import requests
+import logging
+from artbotlib import constants
+
+logger = logging.getLogger(__name__)
+
+
+def process_data(response_data):
```
locriandev, line 8:
```suggestion
def process_data(response_data: dict):
```

artbotlib/pipeline_image_names.py
```diff
@@ -1,5 +1,64 @@
...
+          f"pipeline-image?starting_from={image_type}&name={content_name}&version={version}"
+    logger.debug("URL to server: ", url)
+    response = requests.get(url)
+
+    try:
+        if response.status_code == 200 and response.json().get("status") == "success":
+            payload = process_data(response.json().get("payload"))
+            so.say(payload)
+        else:
+            raise Exception(response.json().get("payload"))
```
locriandev, line 57: If  `response.status_code != 200`, `response.json()` will fail. I'd suggest rewriting this whole block to handle the different issues we might forsee:
- `status code != 200`: something's wrong on art-dash server
- `"status" != "success"`: the server responded with a failure state
- `"payload" not in response.json()`: unexpected payload

E.g.:

```
if response.status_code != 200:
    so.say('server error...')
    return

try:
    json_data = response.json()
except:
    so.say('unexpected payload')
    return

if json_data.get("status") != "success":
   so.say('failure...')
   return

try:
    payload = process_data(json_data["payload"])
    so.say(payload)
except KeyError:
    so.say('...')
```

artbotlib/pipeline_image_util.py
```diff
@@ -567,70 +34,18 @@ def github_distgit_mappings(version: str) -> dict:
...
-            f"image dist-git {distgit_name} definition was not found at {url}")  # If yml file does not exist
-
-    yml_file = yaml.safe_load(response.content)
-    try:
-        return yml_file['distgit']['bundle_component']
-    except KeyError:
-        return None
+    if yml_file.get('for_payload', False):  # Check if the image is in the payload
+        tag = yml_file['name'].split("/")[1]
+        return tag[4:] if tag.startswith("ose-") else tag  # remove 'ose-' if present
```
locriandev, line 51:
```suggestion
        return tag[4:] if tag.startswith("ose-") else tag  # remove 'ose-' if present
    return None
```

locriandev (changes requested):

artbotlib/pipeline_image_names.py
ashwindasr: This will return true if list contains any element right? I was trying to check if the list has 2 or more elements

artbotlib/pipeline_image_names.py
ashwindasr: If response.status_code != 200, then python wont check anything after the subsequent `and` right?

artbotlib/pipeline_image_names.py
ashwindasr: Fixed in https://github.com/openshift-eng/art-bot/pull/72

artbotlib/pipeline_image_names.py
ashwindasr: Fixed in https://github.com/openshift-eng/art-bot/pull/72

artbotlib/pipeline_image_names.py
ashwindasr: Fixed in https://github.com/openshift-eng/art-bot/pull/72

artbotlib/pipeline_image_util.py
ashwindasr: Fixed in https://github.com/openshift-eng/art-bot/pull/72

artbotlib/pipeline_image_names.py
locriandev: Right :) my bad

ashwindasr:
Tested in `art-build-dev` Merging.

PR #69, opened 2023-01-31
Logging for art-bot.py, update logging config

ashwindasr:
- Add logs for art-bot.py
- Update logging config as suggested by @locriandev

### DIFF
--- a/art-bot.py
+++ b/art-bot.py
@@ -3,7 +3,6 @@ import signal
 import click
 import os
 import os.path
-import pprint
 import logging
 import yaml
 import traceback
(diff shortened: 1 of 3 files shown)


art-bot.py
```diff
@@ -92,8 +91,7 @@ def respond(client, event):
         data = event
         web_client = client
 
-        print('\n----------------- DATA -----------------\n')
-        pprint.pprint(data)
```
locriandev, line 96: it might be that in this case we want to keep the `pprint` to have a better looking log

art-bot.py
```diff
@@ -125,8 +123,8 @@ def respond(client, event):
 
         plain_text = extract_plain_text({"data": data}, alt_username)
 
-        print(f'Gating {from_channel}')
```
locriandev, line 128: Maybe these could be logged at `DEBUG` level

ashwindasr: Done!

art-bot.py
ashwindasr: Fixed! The `ppring.pformat` function helped.

PR #72, opened 2023-02-02
Handle json response better

ashwindasr:
Fixes for https://github.com/openshift-eng/art-bot/pull/67

### DIFF
--- a/artbotlib/pipeline_image_names.py
+++ b/artbotlib/pipeline_image_names.py
@@ -21,10 +21,10 @@ def process_data(response_data: dict) -> str:
                    f"<{distgit['brew']['brew_build_url']}|" \
                    f"*{distgit['brew']['brew_package_name']}*>\n"
 
-        if distgit['brew']['bundle_component'] != "None":
+        if distgit['brew']['bundle_component']:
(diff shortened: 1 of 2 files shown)


ashwindasr:
PR for API server: https://github.com/openshift-eng/art-dashboard-server/pull/48

artbotlib/pipeline_image_names.py
```diff
@@ -27,6 +27,10 @@ def process_data(response_data):
         if distgit['brew']['bundle_distgit'] != "None":
```
locriandev, line 27:
```suggestion
        if distgit['brew']['bundle_distgit']:
```

locriandev: Same at line 24 (it won't let me suggest the change)

locriandev (changes requested):

artbotlib/pipeline_image_names.py
ashwindasr: I thought I fixed it... hmm not sure what happened. Fixed in https://github.com/openshift-eng/art-bot/pull/72/commits/c606a1ebeaacab9894581dde0b735cb2c1b0da52

locriandev (approved):

PR #80, opened 2023-02-06
Improve logs in pr_in_build.py

locriandev:


artbotlib/pr_in_build.py
```diff
@@ -196,47 +231,74 @@ def pr_merge_commit(self):
...
         """
+
         successful_builds = self.build_from_commit(BREW_TASK_STATES["Success"])
         if successful_builds:
             self.logger.info("Found successful builds for given PR")
             first_success = successful_builds[0]
             self.so.say(
-                f"First successful build: <{BREW_URL}/buildinfo?buildID={first_success}|{first_success}>. All consecutive builds will include this PR.")
+                f"First successful build: <{BREW_URL}/buildinfo?buildID={first_success}|{first_success}>. "
+                "All consecutive builds will include this PR.")
```
ashwindasr, line 301: We need to merge the two strings right. `so.say` has `**msg_opts`, so would the second string go to that or will both strings still go to the `text` parameter?

locriandev: There's no `,` in between the two, so they're treated as one single string

ashwindasr: Ohh I see okay.

ashwindasr (approved):

PR #87, opened 2023-02-08
[ART-5997] Fix links to RC, display NVR instead of build ID

ashwindasr:
https://issues.redhat.com/browse/ART-5997

### DIFF
--- a/artbotlib/constants.py
+++ b/artbotlib/constants.py
@@ -29,8 +29,7 @@ ART_DASH_API_ROUTE = "http://art-dash-server-aos-art-web.apps.ocp4.prod.psi.redh
 
 RELEASE_CONTROLLER_URL = Template('https://${arch}.ocp.releases.ci.openshift.org')
 
-RELEASE_CONTROLLER_PAGE_URL = Template(
-    "https://${arch}.ocp.releases.ci.openshift.org/releasestream/${type}/release/${name}")
+RELEASE_CONTROLLER_STREAM_PATH = Template("/releasestream/${type}/release/${name}")
 
 PROW_BASE_URL = 'https://prow.ci.openshift.org'
 
--- a/artbotlib/constants.py
+++ b/artbotlib/constants.py
@@ -29,6 +29,9 @@ ART_DASH_API_ROUTE = "http://art-dash-server-aos-art-web.apps.ocp4.prod.psi.redh
 
 RELEASE_CONTROLLER_URL = Template('https://${arch}.ocp.releases.ci.openshift.org')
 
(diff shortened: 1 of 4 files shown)


artbotlib/util.py
```diff
@@ -125,3 +125,14 @@ def ocp_version_from_release_img(release_img: str) -> str:
...
 
     return '.'.join(release_img.split('-')[0].split('.')[:2])
+
+
+def get_build_nvr(build_id):
+    """
+    Get the NVR from the build ID
+    """
+    koji_api = koji_client_session()
+    build = koji_api.getBuild(build_id)
```
locriandev, line 135: We might want to catch exceptions here and notify the user in case of failures

artbotlib/constants.py
```diff
@@ -29,6 +29,9 @@
 
 RELEASE_CONTROLLER_URL = Template('https://${arch}.ocp.releases.ci.openshift.org')
 
+RELEASE_CONTROLLER_PAGE_URL = Template(
```
locriandev, line 32: The root `"https://${arch}.ocp.releases.ci.openshift.org` is already defined in `RELEASE_CONTROLLER_URL`, we might use this to avoid duplication.

locriandev (changes requested):

artbotlib/util.py
ashwindasr: Done!

artbotlib/constants.py
ashwindasr: Done!

artbotlib/util.py
```diff
@@ -125,3 +125,17 @@ def ocp_version_from_release_img(release_img: str) -> str:
...
+    Get the NVR from the build ID
+    """
+    try:
+        koji_api = koji_client_session()
+        build = koji_api.getBuild(build_id)
+        nvr = build['nvr']
+        logger.debug(f"NVR: {nvr}")
+        return nvr
+    except Exception as e:
+        raise BrewNVRNotFound(e)
```
locriandev, line 141: I'd add a logger.error(msg) and optionally a `so.say(msg)` to let the user know what happened, if this is not already handled somewhere else

locriandev (changes requested):

artbotlib/util.py
ashwindasr: I'm handling the exception here and printing the message as well: https://github.com/ashwindasr/art-bot/blob/dfc6eefaa172511bad4d78d7d648167282f62ffc/artbotlib/pr_in_build.py#L301

PR #89, opened 2023-02-13
Fix regex, add GitHub Token

ashwindasr:
Regex was matching only RCs. Fixed to include normal releases as well.

Added GitHub token to prevent rate limiting.


ashwindasr:
Todo:

- [ ] Need to add GitHub token to Deployment configs in `aos-art-web` and `art-build-dev` before merge.

PR #91, opened 2023-02-16
Fix "ERROR There is no current event loop in thread"

locriandev:


locriandev:
Closed in favor of https://github.com/openshift-eng/art-bot/pull/96

PR #94, opened 2023-02-17
Use brew tags defined in erratatool.yml

vfreex:
We are starting to have rhel 9 builds. When query brew, use brew tags defined in ocp-build-data erratatool.yml instead of hardcoded tags.


joepvd (approved):
/lgtm

PR #115, opened 2023-03-20
Fix watch task

locriandev:


artbotlib/help.py
```diff
@@ -21,7 +21,9 @@ def show_help(so):
 * What kernel is used in `release image name or pullspec`?
 
 _*ART release/Brew monitoring:*_
-* Watch `release_url|brew_url|prow_job_url`
```
thegreyd, line 24: I would like to keep the "Watch <url>" interface - it's intuitive and easy to remember
We can remove the ID regex

thegreyd (changes requested):
Would like to keep "watch [url]" for brew.

artbotlib/help.py
locriandev: done

PR #120, opened 2023-03-29
Get data from all pages from GitHub API

ashwindasr:
GitHub API paginates results and the data that we want might not be in the first page that we need. Adding a new function to go through all the pages and get the data.

This update fixes this [issue](https://redhat-internal.slack.com/archives/C03JYDGPX8F/p1680081670976149?thread_ts=1680081654.415989&cid=C03JYDGPX8F). Will work on a consecutive PR to update the other places where would need to go through all the pages to get the results.

### DIFF
--- a/artbotlib/util.py
+++ b/artbotlib/util.py
@@ -2,10 +2,12 @@ import cachetools
 import datetime
 import koji
 import logging
+import functools
+import requests
+import os
 from threading import RLock
 from artbotlib.exceptions import BrewNVRNotFound
 from artbotlib.kerberos import do_kinit
-import functools
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


ashwindasr:
It looks like both the jobs have finished running, but GitHub still shows the brown dot, weird...

artbotlib/util.py
```diff
@@ -139,3 +141,24 @@ def get_build_nvr(build_id):
...
     except Exception as e:
         raise BrewNVRNotFound(e)
+
+
+def github_api_all(url: str):
+    """
+    GitHub API paginates results. This function goes through all the pages and returns everything.
+    """
+    logger.info("Fetching URL using function github_api_all", url)
+    params = {'per_page': 100, 'page': 1}
```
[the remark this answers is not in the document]
ashwindasr: max is 100 per page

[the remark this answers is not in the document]
ashwindasr: hmm I think it also accepts till 1000, but I don't think we have any case in which we go beyond 50, so I kept 100 just to be safe

artbotlib/util.py
```diff
@@ -139,3 +141,24 @@ def get_build_nvr(build_id):
...
+    num_requests = 1  # Guard against infinite loop
+    max_requests = 100
+
+    response = requests.get(url, params=params, headers=header)
+    results = response.json()
+
+    while "next" in response.links.keys() and num_requests <= max_requests:
+        url = response.links['next']['url']
+        response = requests.get(url, headers=header)
+        results += response.json()
```
thegreyd, line 162: So we are merging 2 dicts here? I would prefer that we not use += here, but instead dict.update or 
if github response has a data key. In fetching commits this func works but I don't know how it will 
work for other github apis - so I'll suggest to keep this logic in the above specific func

ashwindasr: Its actually merging two lists, for a similar endpoint like this [one](https://api.github.com/repos/openshift/ovn-kubernetes/commits?sha=release-4.10&since=2023-02-23T15:43:58Z), hmm I'll update the function to thats its clear that its used only for endpoints that returns as a list. Because the ones that return a dict, are usually not paginated.

ashwindasr: updated description in https://github.com/openshift-eng/art-bot/pull/120/commits/62a632942e69372c309ce1ea7e4433077802df24

PR #128, opened 2023-03-31
Sanitize user input

ashwindasr:
Using [shelx.quote](https://docs.python.org/3/library/shlex.html#shlex.quote) to sanitize user input to prevent command Injection vulnerability.

### DIFF
--- a/artbotlib/exectools.py
+++ b/artbotlib/exectools.py
@@ -164,11 +164,9 @@ def cmd_assert(so, cmd, set_env=None, cwd=None, realtime=False):
         raise
 
     if rc:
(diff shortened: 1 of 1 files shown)


ashwindasr:
Looks like GitHub Actions is checking out the base repo instead of the forked repo. Because I tried the synk command locally and it doesn't show any error.

ashwindasr:
Yup, looks like its working on my [branch](https://github.com/ashwindasr/art-bot/actions/runs/4577358344/jobs/8082882099). Have to figure out how to make GitHub Actions checkout the correct code.

PR #130, opened 2023-04-03
snyk action only on base repo

ashwindasr:
Trigger the snyk action only on the base repo. Adding a new condition `github.repository` so that we can check if its the base repository or not.

Updated the authorized users to the [openshift-eng/openshift-team-automated-release](https://github.com/orgs/openshift-eng/teams/openshift-team-automated-release)

### DIFF
--- a/.github/workflows/snyk_security.yaml
+++ b/.github/workflows/snyk_security.yaml
@@ -5,7 +5,7 @@ on: [ push, pull_request_target ]
 jobs:
   snyk:
     runs-on: ubuntu-latest
-    if: contains(fromJSON('["jupierce", "sosiouxme", "thiagoalessio", "joepvd", "thegreyd", "vfreex", "locriandev", "Ximinhan", "ashwindasr"]'), github.actor)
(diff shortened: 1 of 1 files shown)


.github/workflows/snyk_security.yaml
```diff
@@ -5,7 +5,7 @@ on: [ push, pull_request_target ]
 jobs:
   snyk:
     runs-on: ubuntu-latest
-    if: contains(fromJSON('["jupierce", "sosiouxme", "thiagoalessio", "joepvd", "thegreyd", "vfreex", "locriandev", "Ximinhan", "ashwindasr"]'), github.actor)
+    if: contains(fromJSON('["jupierce", "sosiouxme", "thiagoalessio", "joepvd", "thegreyd", "vfreex", "locriandev", "Ximinhan", "ashwindasr"]'), github.actor) && github.repository == 'openshift-eng/art-bot'
```
locriandev, line 8: Do we really care about `github.actor`? I don't like the idea of maintaining this list as years go by :) Besides, I think restricting the owner to be `openshift-eng` would suffice, if there's a way to configure it this way

locriandev (approved):

PR #151, opened 2023-08-29
Add pip_system_certs to requirements

locriandev:
Context here: https://redhat-internal.slack.com/archives/GDBRP5YJH/p1693327136274889


sosiouxme (approved):
/lgtm
i would have expected it in doozer and elliott reqs already but evidently not

PR #154, opened 2023-09-22
Notify user on bot restart

locriandev:


artbotlib/pr_status.py
```diff
@@ -16,14 +16,16 @@ def pr_status(so, user_id, org, repo, pr_id):
...
-        try:
+    variables.active_slack_objects.add(so)
+
+    try:
+        while True:
+            # Timeout after 12 hrs
+            if time.time() - start > constants.ONE_WEEK:
+                so.say(f'PR {pr_url} did not merge after a week.'
+                       f'Giving up...')
+                return
```
ashwindasr, line 27: I think we need to `break` here. Returning will skip the `finally` block, which has the code to remove the `so` object. We had an issue with this then I wrote the nightly job I think...

ashwindasr (changes requested):

ashwindasr (approved):

PR #157, opened 2023-10-12
[ART-7408] Implement fuzzy matching and enhance validation in art-bot

MartinPl0:
Introduced fuzzy matching to suggest commands to users when an unrecognized input is detected. Expanded command-to-regex mapping for improved guidance. Updated 'elliott.py' to include NVR validation and refined error feedback for 'go_nvrs' ' function, ensuring clearer communication and robustness in user interactions. Updated the tests to expect the new output.

### DIFF
--- a/artbotlib/elliott.py
+++ b/artbotlib/elliott.py
@@ -1,4 +1,5 @@
 import logging
+import re
 
 import artbotlib.exectools
 from . import util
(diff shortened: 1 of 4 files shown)


ashwindasr:
I've tested it out locally and looks like its good! 

I'll leave you to merge it when you're online, maybe Monday? So that you'll be there to fix if there are any issues.

artbotlib/elliott.py
```diff
@@ -19,7 +20,16 @@ def image_list(so, advisory_id):
                    filename=f'{advisory_id}.images.txt')
 
 
+def is_valid_nvr(nvr):
+    nvr_regex = re.compile(r'^[a-zA-Z0-9_-]+-[0-9]+\.[0-9]+-[0-9]+$')
```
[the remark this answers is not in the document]
MartinPl0: I will update the regex.

tests/test_regex_mapping.py
```diff
@@ -442,9 +464,16 @@ def test_list_uses_of_rpms(uses_mock):
...
+    query = 'where in 4.10.1 is the ovn rpm used'
+    example_command_valid = 'Where in 4.10 are the rpm1,rpm2,rpm3 rpms used'
+    expected_message = generate_expected_message(example_command_valid)
+    so_mock.should_receive('say').once().with_args(expected_message)
+    map_command_to_regex(so_mock, query, None)
+
     # Invalid - spaces between rpm names
     query = 'where in 4.10 are the rpm1, rpm2, rpm3 rpms used'
-    so_mock.should_receive('say').never()
+    so_mock.should_receive('say').once().with_args(expected_message)
```
MartinPl0, line 476: I have literally no idea why this is causing the so.say to be called twice:
`FAILED tests/test_regex_mapping.py::test_list_uses_of_rpms - flexmock.exceptions.MethodCallError: say() expected to be called exactly 1 time, called 2 times
`
when testing on my local, its working as expected:
```
Enter your command: where in 4.10 are the rpm1, rpm2, rpm3 rpms used
so.say:
I couldn't understand that. For reference, here's an example of a valid command:'Where in 4.10 are the rpm1,rpm2,rpm3 rpms used?' Try following this format or write 'help' to see what I can do!
```

tests/test_regex_mapping.py
```diff
@@ -8,6 +8,8 @@
 
 so = SlackDeveloperOutput()
```
locriandev, line 10: Needs 2 blank lines before and after the function definition

locriandev (commented):
/lgtm

PR #160, opened 2023-10-24
Get newest setuptools

joepvd:


locriandev (commented):
/unhold

If Python 3.11 shows up to be a problem, we can pin 3.8

PR #161, opened 2023-11-14
Add artcommon to install

ashwindasr:
Adding artcommon to Dockerfile install

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -3,8 +3,8 @@ LABEL name="art-bot" \
   description="art-bot container image" \
   maintainer="OpenShift Automated Release Tooling (ART) Team <aos-team-art@redhat.com>"
 
-# the build will need to run inside the firewall to access internal resources.
-# install Red Hat IT Root CA and RCM repos
+# This build will need to be run inside the firewall to access internal resources.
+# Install Red Hat IT Root CA and RCM repos, runtime dependencies, and upgrade pip
 RUN curl -o /etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt --fail -L \
     https://certs.corp.redhat.com/certs/2015-IT-Root-CA.pem \
  && curl -o /etc/pki/ca-trust/source/anchors/2022-IT-Root-CA.pem --fail -L \
@@ -16,15 +16,17 @@ RUN curl -o /etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt --fail -L \
     krb5-workstation git rsync \
     python3 python3-certifi python3-rpm python3-rhmsg \
     # development dependencies
-    gcc krb5-devel python3-devel python3-pip python3-setuptools \
+    gcc krb5-devel python3-devel python3-pip \
     # other tools
     bash-completion vim tmux wget curl iputils procps-ng psmisc net-tools iproute \
     # install brewkoji
     koji brewkoji \
- && dnf clean all
+ && dnf clean all \
+    # upgrade pip
+ && pip3 install --upgrade pip
 
+# Install OpenShift Client
 ARG OC_VERSION=candidate
-# include oc client
 RUN wget -O /tmp/openshift-client-linux-"$OC_VERSION".tar.gz https://mirror.openshift.com/pub/openshift-v4/clients/ocp/"$OC_VERSION"/openshift-client-linux.tar.gz \
   && tar -C /usr/local/bin -xzf  /tmp/openshift-client-linux-"$OC_VERSION".tar.gz oc kubectl \
   && rm /tmp/openshift-client-linux-"$OC_VERSION".tar.gz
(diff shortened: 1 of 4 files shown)


container/Dockerfile
```diff
@@ -38,35 +38,41 @@ ARG USER_GID=$USER_UID
...
  && chmod 0440 /etc/sudoers.d/"$USERNAME"
 
-USER "$USER_UID"
+# Set work directory
 WORKDIR /workspaces/art-bot
 
-# install dependencies (allow even openshift's random user to see)
-ENV PATH=/home/"$USERNAME"/.local/bin:/home/"$USERNAME"/bin:"$PATH"
+# Switch to the root user for installations
+USER 0
```
ashwindasr, line 53: I've been reading about root users in Dockerfiles recently, and what I understood that by default it is always root until we specify a non root user. So I think this can be removed. Can you check once to see if its working?

container/Dockerfile
```diff
@@ -38,35 +38,41 @@ ARG USER_GID=$USER_UID
...
+
+# Install Python, upgrade pip, and Git, and clone art-tools
+RUN dnf install -y git python3.8 python3-pip \
```
ashwindasr, line 56: We are already installing python and pip above, so this line is redundant

container/Dockerfile
```diff
@@ -38,35 +38,41 @@ ARG USER_GID=$USER_UID
...
+ && pip3 install --upgrade pip \
```
ashwindasr, line 57: Lets move this upgrade pip to after pip is installed, on Line 24 so that its visible that we did it once and we don't need to do it later.

So adding `&& cd art-tools && ./install.sh` should be sufficient for our case

container/Dockerfile
MartinPl0: It works!

container/Dockerfile
MartinPl0: Thank you.

container/Dockerfile
```diff
@@ -35,38 +37,36 @@ ARG USERNAME=dev
...
 COPY . /tmp/art-bot
-USER 0
 RUN cp -r /tmp/art-bot/{artbotlib,art-bot.py,art_bot_dev.py} . \
  && cp /tmp/art-bot/container/doozer-settings.yaml /home/"$USERNAME"/.config/doozer/settings.yaml \
  && cp /tmp/art-bot/container/elliott-settings.yaml /home/"$USERNAME"/.config/elliott/settings.yaml \
  && rm -rf /tmp/art-bot
+
+# Return to non-root user and final work directory
 USER "$USER_UID"
+WORKDIR /workspaces/art-bot
```
ashwindasr, line 72: I don't think we need to reset `WORKDIR` since we already set it above

container/Dockerfile.latest
```diff
@@ -11,19 +11,22 @@ ARG USER_GID=$USER_UID
 
 USER 0
 
-# install dependencies (allow even openshift's random user to see)
+WORKDIR /art-tools
```
ashwindasr, line 14: Lets not set `WORKDIR` here as well and operate from root, like we did for `Dockerfile.base`

container/Dockerfile.latest
```diff
@@ -11,19 +11,22 @@ ARG USER_GID=$USER_UID
...
+RUN git pull
+RUN ./install.sh
+
+# Install dependencies from requirements.txt
 COPY requirements.txt ./
-RUN umask a+rx && pip3 install --upgrade \
-    "git+https://github.com/openshift-eng/art-tools.git#egg=rh-doozer&subdirectory=doozer" \
-    "git+https://github.com/openshift-eng/art-tools.git#egg=rh-elliott&subdirectory=elliott" \
-    -r ./requirements.txt
+RUN pip3 install --upgrade -r requirements.txt
```
ashwindasr, line 20: Hmm I believe `pip` updates relatively often than others. Lets add a line to update that before the first pip command in this file

container/Dockerfile.latest
```diff
@@ -11,19 +11,22 @@ ARG USER_GID=$USER_UID
...
+WORKDIR /art-tools
+RUN git pull
+RUN ./install.sh
+
+# Install dependencies from requirements.txt
 COPY requirements.txt ./
-RUN umask a+rx && pip3 install --upgrade \
-    "git+https://github.com/openshift-eng/art-tools.git#egg=rh-doozer&subdirectory=doozer" \
-    "git+https://github.com/openshift-eng/art-tools.git#egg=rh-elliott&subdirectory=elliott" \
-    -r ./requirements.txt
```
ashwindasr, line 19: We need these, since every time we art-tools updates, we need to update this as well. You can follow a similar pattern to `RUN umask a+rx && pip3 install --user --upgrade rh-doozer rh-elliott rh-artcommon -r ./requirements.txt`

ashwindasr:
@MartinPl0 take a look at https://github.com/openshift-eng/art-bot/pull/161/commits/7b4e9c41d6be06414b791d48709bcedfc9de2e2e when you get a chance. Things to note:
- Sometimes I can be wrong about the `USER 0` part, it caused an installation error in Dockerfile.latest, when I tried to test. So always make sure that you are updating all places a piece of code is used and make sure it works 
- `./` is different from `/`. `/art-tools` means an art-tools folder in the root directory, while `./art-tools` means one in the current dir. We need the latter

ashwindasr:
Tested the builds in my hackspace, so I think it should work. Merging

container/Dockerfile.latest
```diff
@@ -3,27 +3,37 @@ FROM art-bot:base
...
+
 # use same non-root user from the initial install
 ARG USERNAME=dev
 # On Linux, replace with your actual UID, GID if not the default 1000
 ARG USER_UID=1000
 ARG USER_GID=$USER_UID
 
+WORKDIR /workspaces/art-bot
+
+RUN git config --global --add safe.directory /workspaces/art-bot/art-tools
```
ashwindasr, line 17: This can be removed if we use USER 0, need to check

PR #163, opened 2023-11-28
[ART-8304] Clone art-tools into working directory and clean up post-install

MartinPl0:
This commit changes the Dockerfiles to clone art-tools into the current working directory instead of a fixed path. It also ensures that the art-tools directory is removed after the installation process to keep the Docker image clean. This approach aligns with best practices for container filesystem organization and avoids potential issues with OpenShift's random user ID assignment.

### DIFF
--- a/container/Dockerfile.latest
+++ b/container/Dockerfile.latest
@@ -12,16 +12,16 @@ ARG USERNAME=dev
 ARG USER_UID=1000
 ARG USER_GID=$USER_UID
 
-WORKDIR /workspaces/art-bot
-
 RUN git config --global --add safe.directory /workspaces/art-bot/art-tools
 
-# Looks like we need a USER 0 here
+# Switch to root user for git operations
 USER 0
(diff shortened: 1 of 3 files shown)


container/Dockerfile.latest
```diff
@@ -16,12 +16,15 @@ WORKDIR /workspaces/art-bot
...
+# Switch to root user for git operations
 USER 0
-# Clone art-tools and run install.sh script
-RUN cd art-tools \
- && git pull \
- && ./install.sh
+# Update art-tools and run the install script
+RUN git clone https://github.com/openshift-eng/art-tools.git \
+    && cd art-tools \
+    && git pull \
```
ashwindasr, line 24: Since we are cloning art-tools, `git pull` would not be necessary

ashwindasr (changes requested):

ashwindasr:
Also can consider removing `WORKDIR` in Dockerfile.update, after testing to see if it works

PR #164, opened 2024-01-02
ART-8559 Use payload_name when determining imagestream tag

locriandev:
Ref. https://issues.redhat.com/browse/ART-8559


thiagoalessio (approved):
/lgtm

PR #167, opened 2024-01-16
ART-8680: remove catalogs from art-bot query

mbiarnes:
since  @art-bot : **what catalogs are associated with `release-tag` was returning**
wrongs URLs we decided to remove the possibilities to generate this command with `catalogs`


ashwindasr:
/hold

This is fixable

PR #174, opened 2024-02-20
fixed art-bot query using catalogs

mbiarnes:


artbotlib/brew_list.py
```diff
@@ -124,13 +122,38 @@ async def get_tag_specs(so, tag_spec, data_type, sem) -> str:
...
+            if suffix in component_name:
+                catalog_name = component_name.rstrip('container')
+                catalog_name = catalog_name.rstrip("-")
+
+            version = component_version.strip('v')
+            version = version[:4]
+            data_info = catalog_api_query(so, image_type, catalog_name, version)
+            logger.info(f"RECEIVED CATALOG_DATA: {data_info}")
+
+            result += f'https://catalog.redhat.com/software/containers/{data_info[0]}/{data_info[1]}'
```
ashwindasr, line 137: Lets keep the catalog base URL in `constants.py`

ashwindasr (changes requested):

mbiarnes:
@ashwindasr now it is working also if we have multiple CDNs or deliveries. Please test it.

artbotlib/brew_list.py
```diff
@@ -120,18 +119,62 @@ async def get_tag_specs(so, tag_spec, data_type, sem) -> str:
...
+                catalog_name = catalog_name.rstrip("-")
+            version = component_version.strip('v')
+            version = version[:4]
+            data_info = catalog_api_query(so, image_type, catalog_name, version)
+
+            try:
+                for item in data_info:
+                    cat_id, cat_name = item
+                    result += f"{constants.CATALOG_URL}/{cat_name}/{cat_id}\n"
+            except Exception as e:
```
locriandev, line 141: What exception do you expect here? I think we can be more specific and restrict the `except` block to the expected exception only

artbotlib/brew_list.py
```diff
@@ -120,18 +119,62 @@ async def get_tag_specs(so, tag_spec, data_type, sem) -> str:
...
         elif data_type.startswith('catalog'):
-            result += f'{component_rhcc_url}'
+            image_type = 'distgit'
+            suffix = "-container"
+            catalog_name = ""
+            if suffix in component_name:
+                catalog_name = component_name.rstrip('container')
+                catalog_name = catalog_name.rstrip("-")
+            version = component_version.strip('v')
+            version = version[:4]
```
locriandev, line 134: We might benefit from a comment that explains the format you're getting, and why you're taking only the first 4 characters

artbotlib/brew_list.py
```diff
@@ -120,18 +119,62 @@ async def get_tag_specs(so, tag_spec, data_type, sem) -> str:
...
+            data_info = catalog_api_query(so, image_type, catalog_name, version)
+
+            try:
+                for item in data_info:
+                    cat_id, cat_name = item
+                    result += f"{constants.CATALOG_URL}/{cat_name}/{cat_id}\n"
+            except Exception as e:
+                # Handle any exceptions that might occur within the try block
+                print("Payload is empty:", e)
```
locriandev, line 143:
```suggestion
                logger.warning("Payload is empty:", e)
```

artbotlib/brew_list.py
```diff
@@ -120,18 +119,62 @@ async def get_tag_specs(so, tag_spec, data_type, sem) -> str:
...
+
+    if response.status_code != 200:
+        logger.error(f"API Server error. Status code: {response.status_code}")
+        so.say("API server error. Contact ART Team.")
+        so.monitoring_say("ERROR: API server error.")
+        return
+
+    try:
+        catalogs_info = response.json()
+    except Exception as e:
```
locriandev, line 166: Again, if the response payload cannot be parsed as JSON, we'll get a specific Exception class and that's what we should be catching here

locriandev (changes requested):

mbiarnes:
@ashwindasr 
there are still two things I don't like:

1. when we have i.e. the component_name: ose-cluster-policy-controller-container
there does not exist a distgit for this repo, but for cluster-policy-controller.
When it is easy to strip `-container` it is not clear how to look if ose is needed or not.

2. When accessing an URL like https://art-dash-server-art-dashboard-server.apps.artc2023.pc3z.p1.openshiftapps.com/api/v1/pipeline-image?starting_from=distgit&name=golang-github-prometheus-alertmanager&version=4.16 it takes a lot of time (also when you browse to this URL). 
So the command for art -bot: What catalogs are associated with <release-tag> takes very long! I don't think we can accelerate this.

ashwindasr:
@mbiarnes 

1. We store the brew package name in the image metadata (ocp-build-data config file) itself, so doozer runtime will have the mapping (since the distgit key is the file name)
2. Yes right now its not optimized, but it has the potential to be cached on the server side, if you are looking to optimize it

mbiarnes:
what catalog are associated with <release> works.
Before it could fetch from the json file ['config']['config']['label'] the url, but this doesn't work any more.
The resulting catalog url has to be constructed by accessing urls like:
https://art-dash-server-art-dashboard-server.apps.artc2023.pc3z.p1.openshiftapps.com/api/v1/pipeline-image?starting_from=distgit&name=ose-alibaba-cloud-controller-manager&version=4.16
and retrieving from the json answer
delivery_repo_id and delivery_repo_name 
The access to https://art-dash-server-art-dashboard-server.apps.artc2023.pc3z.p1.openshiftapps.com/api/v1/pipeline-image?starting_from=distgit&name=>calatog_name>&version=<MAJ.MIN) takes a long amount of time

PR #177, opened 2024-04-08
fix nightly url

Ximinhan:
fix nightly pullspec url prefix

### DIFF
--- a/artbotlib/brew_list.py
+++ b/artbotlib/brew_list.py
@@ -14,7 +14,7 @@ import yaml
 import artbotlib.exectools
 
 from . import util
-from .constants import RHCOS_BASE_URL
+from .constants import NIGHTLY_REGISTRY, QUAY_REGISTRY
 from .rhcos import RHCOSBuildInfo
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


artbotlib/brew_list.py
```diff
@@ -143,7 +143,7 @@ def list_component_data_for_release_tag(so, data_type, release_tag):
     so.say('Let me look into that. It may take a minute...')
 
     if 'nightly-' in release_tag:
-        repo_url = 'registry.svc.ci.openshift.org/ocp/release'
+        repo_url = 'registry.ci.openshift.org/ocp/release'
```
locriandev, line 146: Can we make it a constant? It's already being used in many parts of the code, but hardcoded as a string

Ximinhan: I only saw it also used once in rhcos lib, let's improve it later when we got more use case

Ximinhan: updated in https://github.com/openshift-eng/art-bot/pull/177/commits/af075f51a57cabed5273b6a39d5e43d4d2953dce

artbotlib/constants.py
```diff
@@ -35,6 +35,10 @@
 
 PROW_BASE_URL = 'https://prow.ci.openshift.org'
 
+NIGHTLY_REGISTRY = 'registry.ci.openshift.org/ocp/release'
```
locriandev, line 38: We should use `registry.ci.openshift.org` instead, which is referenced in many other places, eg:
- pyartcd/constant.py:13
- doozerlib/cli/images_streams.py:596
and a few moew

Ximinhan: this is art-bot repo

PR #179, opened 2024-04-15
ART-6684 Get latest nightly status

adobes1:


artbotlib/nightly_color.py
```diff
@@ -62,6 +62,43 @@ def get_failed_jobs(release_url, release_browser) -> str:
...
+def get_latest_nightly_name(release_brower, release_stream) -> str:
+    url = f"https://{release_brower}.ocp.releases.ci.openshift.org/api/v1/releasestream/{release_stream}/tags"
+    logger.info('Fetching URL %s', url)
+
+    response = requests.get(url)  # query API to get list of nightly tags
+    status_code = response.status_code
+    if status_code != 200:
+        logger.error('Server responded with status code %s', status_code)
+
+    return response.json()['tags'][0]['name']
```
joepvd, line 74: This will fail, especially when status_code is not 200.

artbotlib/nightly_color.py
```diff
@@ -62,6 +62,43 @@ def get_failed_jobs(release_url, release_browser) -> str:
     return payload
 
 
+def get_latest_nightly_name(release_brower, release_stream) -> str:
+    url = f"https://{release_brower}.ocp.releases.ci.openshift.org/api/v1/releasestream/{release_stream}/tags"
```
joepvd, line 66: Typo. Maybe call it `arch`?

artbotlib/nightly_color.py
```diff
@@ -62,6 +62,43 @@ def get_failed_jobs(release_url, release_browser) -> str:
...
+    logger.info('Fetching URL %s', url)
+
+    response = requests.get(url)  # query API to get list of nightly tags
+    status_code = response.status_code
+    if status_code != 200:
```
joepvd, line 71: variable name is not needed, can call `response.status_code` directly here

artbotlib/regex_mapping.py
```diff
@@ -307,6 +307,13 @@ def map_command_to_regex(so, plain_text, user_id):
             "user_id": True,
             "example": "Watch https://amd64.ocp.releases.ci.openshift.org/releasestream/4.13.0-0.ci/release/4.13.0-0.ci-2022-12-19-111818"
         },
+        {
+            "regex": r"^Alert ?(if|when|on)? latest ? (?P<release_browser>[\w]+)? ?(?P<version>\d+.\d+) ?(stops being blue|fails|is rejected|is red|is accepted|is green)?$",
+            "flag": re.I,
+            "function": latest_nightly_color_status,
+            "user_id": True,
+            "example": "Alert when latest s390x 4.16 is accepted"
```
Ximinhan, line 315: I prefer the pattern "watch latest 4.16", no need to match those extra words

artbotlib/nightly_color.py
```diff
@@ -62,6 +62,43 @@ def get_failed_jobs(release_url, release_browser) -> str:
...
+    :param user_id: User ID of the person who invoked ART-Bot
+    :param version: OCP version eg: 4.15/4.16
+    :param release_browser: release browser eg: ppc64le/arm64/s390x/amd64
+    :return: None
+    """
+
+    if release_browser is None: # Assume amd64 if no arch given
+        release_browser = "amd64"
+
+    if release_browser == "amd64":
```
Ximinhan, line 91: we usually only care about amd64, other arch like s390x don't have test jobs so won't need to wait.

artbotlib/nightly_color.py
adobes1: This should be fixed now

artbotlib/nightly_color.py
adobes1: As per Ximin's comment, I have removed the option to use architectures other than amd64

artbotlib/nightly_color.py
adobes1: Thank you, updated

artbotlib/regex_mapping.py
```diff
@@ -307,6 +307,13 @@ def map_command_to_regex(so, plain_text, user_id):
             "user_id": True,
             "example": "Watch https://amd64.ocp.releases.ci.openshift.org/releasestream/4.13.0-0.ci/release/4.13.0-0.ci-2022-12-19-111818"
         },
+        {
+            "regex": r"^Alert ?(if|when|on)? latest ? (?P<release_browser>[\w]+)? ?(?P<version>\d+.\d+) ?(stops being blue|fails|is rejected|is red|is accepted|is green)?$",
```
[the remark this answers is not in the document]
adobes1: I have modified the pattern according to Ximin's suggestion, which simplified the regex quite a bit. Hopefully it is correct now

artbotlib/regex_mapping.py
adobes1: Fixed

artbotlib/nightly_color.py
adobes1: I see, I have updated it accordingly

PR #181, opened 2024-04-24
Fix art-bot what kernel is in rhcos cmd

thegreyd:
When Release Controller cannot load changelog for a nightly, we cannot fetch rhcos build id.
So instead of fetching it via RC api, fetch it from release image. 

Test:
```
./art_bot_dev.py
Enter your command: What kernel is in 4.16.0-0.nightly-2024-04-24-091844
```

### DIFF
--- a/artbotlib/kernel_info.py
+++ b/artbotlib/kernel_info.py
@@ -68,7 +68,7 @@ class KernelInfo:
         pullspec, _ = buildinfo.get_img_pullspec(self.release_img)
         rhcos_build_id = await rhcos.get_rhcos_build_id_from_pullspec(pullspec)
         if not rhcos_build_id:
-            self.logger.warning('Failed to fetch RHCOS info for %s', self.release_img)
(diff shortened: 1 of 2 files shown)


artbotlib/kernel_info.py
```diff
@@ -81,13 +80,6 @@ async def rhcos_kernel_info(self):
         kernel_core = [pkg for pkg in pkg_list if 'kernel-core' in pkg][0]
         rpms.append(f'kernel-core.{".".join(kernel_core[2:])}')
 
-        # Get kernel-rt-core from build labels, if available
-        build_info, pullspec, _ = await buildinfo.get_image_info(
-            self.so, 'machine-os-content', self.release_img)
-        labels = build_info['config']['config']['Labels']
-        if 'com.coreos.rpm.kernel-rt-core' in labels:
```
thegreyd, line 88: These no longer exist

PR #183, opened 2024-05-23
Allow listing rpms for existing releases

jupierce:


sosiouxme:
to be clear, i think this has the effect of including the RPMs ART released... but not every RPM used in every image for a release.

ashwindasr (approved):

PR #188, opened 2024-07-03
ART-10035: use pr target branch when check commits

Ximinhan:
when checking commits for a pr, the static branch "release-{self.version}" name value is not always like this. use pr target branch as a source of truth


artbotlib/pr_in_build.py
```diff
@@ -208,7 +208,9 @@ def pr_merge_commit(self):
         try:
             sha = json_data["merge_commit_sha"]
             self.logger.info('Found merge commit SHA: %s', sha)
-            return sha
+            branch = json_data["base"]["ref"]
+            elf.logger.info('Merge request branch: %s', branch)
```
mbiarnes, line 212: typo - an `s` missing

PR #190, opened 2024-07-03
ART-9985: art-bot notifies of all builds for a PR

mbiarnes:


artbotlib/pr_in_build.py
```diff
@@ -248,8 +248,18 @@ def build_from_commit(self, task_state):
...
+                    image_name = b["build_0_name"]
+                    build_id = b["build_0_id"]
+                    if image_name not in builds_by_image:
+                        builds_by_image[image_name] = []
+                    builds_by_image[image_name].append(build_id)
+
+                # Extract lists of build IDs using list comprehension
+                build_id_list = [build_id for build_id in builds_by_image.values()]
+                sorted_earliest_build = [sorted(sublist)[0] for sublist in build_id_list]
+                return (sorted_earliest_build)
```
Ximinhan, line 262: How is this different from the original implementation?

artbotlib/pr_in_build.py
```diff
@@ -261,24 +271,24 @@ def find_builds(self):
...
-            except BrewNVRNotFound as e:
-                self.logger.error(f"Could not find NVR from build: {e}")
-                self.so.say("Could not find NVR from build")
-                return
-            self.so.say(
-                f"First successful build: <{BREW_URL}/buildinfo?buildID={first_success}|{nvr}>. All consecutive builds will include this PR.")
+            nvr = None
+            for build in successful_builds:
+                try:
+                    nvr = util.get_build_nvr(build)
```
Ximinhan, line 277: I think the build nvr also in the the db, you can use the build_0_nvr from db so don't need to get from brew again

artbotlib/pr_in_build.py
Ximinhan: seems you want to get a list of build_id for each image, suggest change

max_values = {}  
for key, value in [(b["build_0_name"], b["build_0_id"]) for b in builds]:  
    if key not in max_values or value > max_values[key]:  
        max_values[key] = value 
result = [value for key, value in max_values.items()]
return result

mbiarnes: the difference to original implementation: 
see comment in [ART-9985](https://issues.redhat.com/browse/ART-9985)

artbotlib/pr_in_build.py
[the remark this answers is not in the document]
mbiarnes: .. this is more or less like it was before!
Before these changes it was retrieved the nvr by `util.get_build_nvr`.
I don't now if we should change this, since this ever worked this way.

artbotlib/pr_in_build.py
mbiarnes: see [refactoring function to get build_ids for each image](https://github.com/openshift-eng/art-bot/pull/190/commits/35b5653031c84063e2d8edf065665bfbb03d3814)

mbiarnes: > seems you want to get a list of build_id for each image, suggest change
> 
> max_values = {} for key, value in [(b["build_0_name"], b["build_0_id"]) for b in builds]: if key not in max_values or value > max_values[key]: max_values[key] = value result = [value for key, value in max_values.items()] return result

see PR refactoring function to get build_ids for each image](https://github.com/openshift-eng/art-bot/pull/190/commits/35b5653031c84063e2d8edf065665bfbb03d3814)

mbiarnes: > How is this different from the original implementation?

the difference to original implementation:
see comment in [ART-9985](https://issues.redhat.com/browse/ART-9985)

artbotlib/pr_in_build.py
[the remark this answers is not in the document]
mbiarnes: > I think the build nvr also in the the db, you can use the build_0_nvr from db so don't need to get from brew again



.. this is more or less like it was before!
Before these changes it was retrieved the `nvr by util.get_build_nvr`.
I don't now if we should change this, since this ever worked this way.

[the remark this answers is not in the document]
Ximinhan: yes, i mean its a place we can improve, I am fine we have another pr to do that

artbotlib/pr_in_build.py
```diff
@@ -248,8 +248,12 @@ def build_from_commit(self, task_state):
             if response_data and count > 0:
                 self.logger.info('Found %s builds from commit %s', count, commit)
                 builds = response_data["results"]
-                build_ids = [x["build_0_id"] for x in builds]
-                return sorted(build_ids)
+                max_values = {}
+                for key, value in [(b["build_0_name"], b["build_0_id"]) for b in builds]:
+                    if key not in max_values or value > max_values[key]:
```
thegreyd, line 253:
```suggestion
                    if key not in max_values or value < max_values[key]:
```

artbotlib/pr_in_build.py
```diff
@@ -248,8 +248,12 @@ def build_from_commit(self, task_state):
             if response_data and count > 0:
                 self.logger.info('Found %s builds from commit %s', count, commit)
                 builds = response_data["results"]
-                build_ids = [x["build_0_id"] for x in builds]
-                return sorted(build_ids)
+                max_values = {}
+                for key, value in [(b["build_0_name"], b["build_0_id"]) for b in builds]:
```
thegreyd, line 252:
```suggestion
                for key, value in [(b["build_0_name"], int(b["build_0_id"])) for b in builds]:
```

thegreyd: let's be descriptive with our var names, it will be clear on what we are donig

artbotlib/pr_in_build.py
Ximinhan: oh, right, maybe the value name `max_values` also changed to `min_values`

artbotlib/pr_in_build.py
mbiarnes: Done: https://github.com/openshift-eng/art-bot/pull/190/commits/572dab3aad2815a752fb90834f4e96f1844f013d

artbotlib/pr_in_build.py
```diff
@@ -248,8 +248,12 @@ def build_from_commit(self, task_state):
...
+                    if key not in max_values or value > max_values[key]:
+                        max_values[key] = value
+                result = [value for key, value in max_values.items()]
+                return result
```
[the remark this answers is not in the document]
mbiarnes: Done: https://github.com/openshift-eng/art-bot/pull/190/commits/572dab3aad2815a752fb90834f4e96f1844f013d

artbotlib/pr_in_build.py
```diff
@@ -248,8 +248,15 @@ def build_from_commit(self, task_state):
             if response_data and count > 0:
                 self.logger.info('Found %s builds from commit %s', count, commit)
                 builds = response_data["results"]
-                build_ids = [x["build_0_id"] for x in builds]
-                return sorted(build_ids)
+                build_ids = [(x["build_0_id"], x["build_0_name"]) for x in builds]
+                self.logger.info(f' *** these are the build_ids {build_ids} ***')
+
+                builds = response_data["results"]
+                max_values = {}
```
thegreyd, line 255: the name max_values doesn't make sense here, let's change it to something else

```suggestion
                build_for_image = {}
```

mbiarnes: Done

artbotlib/pr_in_build.py
```diff
@@ -248,8 +248,15 @@ def build_from_commit(self, task_state):
             if response_data and count > 0:
                 self.logger.info('Found %s builds from commit %s', count, commit)
                 builds = response_data["results"]
-                build_ids = [x["build_0_id"] for x in builds]
-                return sorted(build_ids)
+                build_ids = [(x["build_0_id"], x["build_0_name"]) for x in builds]
+                self.logger.info(f' *** these are the build_ids {build_ids} ***')
```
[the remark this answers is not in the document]
mbiarnes: leftover - removed!

PR #194, opened 2024-07-17
show only images which should be in payload

mbiarnes:
excluded images that are not for payload

### DIFF
--- a/artbotlib/pr_in_build.py
+++ b/artbotlib/pr_in_build.py
@@ -6,6 +6,7 @@ import os
 from collections.abc import Iterable
 
 import requests
+import yaml
 
 import artbotlib.exectools
 from artbotlib import util, pipeline_image_util
@@ -236,6 +237,19 @@ class PrInfo:
 
         return response.json()
 
+    def is_image_for_release(self, image_name):
+        """
+        Exclude images that are not in payloads
+        """
+        url = f'https://raw.githubusercontent.com/openshift-eng/ocp-build-data/openshift-{self.version}/images/{image_name}.yml'
+        response = requests.get(url)
+        if response.status_code == 200:
+            yaml_content = yaml.safe_load(response.text)
+            # Check key value for for_relaese
+            return yaml_content.get('for_release', True)
+        else:
+            self.logger.info(f'response.status_code = {response.status_code}')
+
     def build_from_commit(self, task_state):
         """
         Function to get all the build ids associated with a list of commits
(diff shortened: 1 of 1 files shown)


artbotlib/pr_in_build.py
```diff
@@ -236,6 +237,23 @@ def get_builds_from_db(self, commit, task_state):
...
+    def exclude_images(self, image_name):
+        """
+        Exclude images that are not in payloads
+        """
+        url = f'https://raw.githubusercontent.com/openshift-eng/ocp-build-data/openshift-{self.version}/images/{image_name}.yml'
+        response = requests.get(url)
+        if response.status_code == 200:
+            yaml_content = yaml.safe_load(response.text)
+            # Check key value for for_payload
+            value = yaml_content.get('for_payload', True)
```
thegreyd, line 249: ah so this should be ('for_release', True)

operators will have for_payload set to False but we still want to include them here. So we should only look at for_release field

artbotlib/pr_in_build.py
```diff
@@ -236,6 +237,23 @@ def get_builds_from_db(self, commit, task_state):
 
         return response.json()
 
+    def exclude_images(self, image_name):
```
thegreyd, line 240: I would recommend the function name, something like "is_image_for_release(self, image_name)"

artbotlib/pr_in_build.py
```diff
@@ -249,8 +267,9 @@ def build_from_commit(self, task_state):
                 self.logger.info('Found %s builds from commit %s', count, commit)
                 builds = response_data["results"]
                 build_for_image = {}
-                for image_name, build_id in [(b["build_0_name"], int(b["build_0_id"])) for b in builds]:
-                    if image_name not in build_for_image or build_id < build_for_image[image_name]:
+                for image_name, build_id, dg_name in [(b["build_0_name"], int(b["build_0_id"]), b["dg_name"]) for b in builds]:
```
thegreyd, line 270: so i think the distgit name can sometimes be different from the image name we have in yaml.. 
can you check? and if that's the case, can we use something else here?

mbiarnes: build_0_name without -containers should be == dg_name - but this is not every time the case.
But the builds have build_0_name and dg_name and are based on build_0_id. So I think here is no peroblem. At least 
pr info https://github.com/openshift/ovn-kubernetes/pull/2209 in 4.16
is working fine and here is an example that distgit name is different from image_name.

artbotlib/pr_in_build.py
mbiarnes: DONE with return yaml_content.get('for_release', True)

PR #195, opened 2024-07-30
[ART-10106] get build nvr from db

Ximinhan:
when query the build, the build nvr should already in the db so no need to get the build_id from db then query brew to get the nvr again.


mbiarnes:
tested locally and it worked

PR #221, opened 2025-09-10
Treat master/main as equivalent

thegreyd:


ashwindasr (approved):

ashwindasr:
Might need to manually trigger a build for art-bot.

smeeio is not working properly right now I think

